PR #3, opened 2020-09-10
fix bazel; usage comment

jayfurmanek:
Fix for bazel - left over from internal where we had multiple bazel builds.

Also fix a old usage comment in `build_feedstock.py`


bethune-bryant (approved):
Lgtm

bnemanich (approved):

PR #10, opened 2020-09-10
Fix bug with the --repository_folder argument

bnemanich:
There was a bug with the repository_folder argument. I added a fix for it, as well as a test.


bethune-bryant (approved):

jayfurmanek (approved):
Yup. Good catch.

PR #11, opened 2020-09-10
Add config linting.

### LINKED ISSUE #9: Add static analysis for config files.
Since we're not building the environments on PRs,we should add some validation tests to make sure that nothing gets broken.

bethune-bryant:
Closes #9

### DIFF
--- a/builder/build_env.py
+++ b/builder/build_env.py
@@ -124,7 +124,7 @@ def make_hash(d):
 
 def validate_config_file(env_file, variants):
     '''Perform some validation on the environment file after loading it.'''
-    possible_keys = {'imported_envs', 'channels', 'packages'}
+    possible_keys = {'imported_envs', 'channels', 'packages', 'git_tag_for_env', 'git_tag'}
     try:
         meta_obj = conda_build.metadata.MetaData(env_file, variant=variants)
         if not ("packages" in meta_obj.meta.keys() or "imported_envs" in meta_obj.meta.keys()):
--- a/builder/build_env.py
+++ b/builder/build_env.py
@@ -122,15 +122,36 @@ repositories will be downloaded to the current working directory.""")
 def make_hash(d):
     return hash(str(d))
 
+def validate_config_file(env_file, variants):
+    '''Perform some validation on the environment file after loading it.'''
+    possible_keys = {'imported_envs', 'channels', 'packages'}
+    try:
+        meta_obj = conda_build.metadata.MetaData(env_file, variant=variants)
+        if not ("packages" in meta_obj.meta.keys() or "imported_envs" in meta_obj.meta.keys()):
+            raise Exception("Content Error!", "An environment file needs to specify packages or import another environment file.")
+        for key in meta_obj.meta.keys():
+            if not key in possible_keys:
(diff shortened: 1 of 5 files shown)


tests/envlint.py
```diff
@@ -0,0 +1,41 @@
...
+def main(arg_strings=None):
+    parser = make_parser()
+    args = parser.parse_args(arg_strings)
+
+    retval = 0
+    for env_file in args.env_files:
+        try:
+            meta_obj = conda_build.metadata.MetaData(env_file, variant=variants)
+            if not ("packages" in meta_obj.meta.keys() or "imported_envs" in meta_obj.meta.keys()):
+                raise Exception("Content Error!", "An environment file needs to specify packages or import another environment file.")
```
bethune-bryant, line 33: @bnemanich 
Does it make sense to have this requirement?

bethune-bryant: Are there any others that I should add?

bnemanich: I think this one makes sense. Another possible one is to check for any unknown keys. Those probably won't cause an error, but it could catch things like `packages` being spelt `pckages`.

bethune-bryant: Ooooh, that's a good one.
I'll add that.
Are the only possibilities:
```
imported_envs
channels
packages
```

bnemanich: Also, we might want this to be part of the actual build tool as well, to check env files before they are loaded. Maybe the heart of this could be moved to the build directory, and this file would just import the function.

bethune-bryant: That's a good idea.
I'll make that change.

bethune-bryant: @bnemanich 
I made this change. Let me know what you think.

bethune-bryant:
@bnemanich 
What do you think of these changes?

builder/build_env.py
```diff
@@ -122,15 +122,36 @@ def make_parser():
 def make_hash(d):
     return hash(str(d))
 
+def validate_config_file(env_file, variants):
+    '''Perform some validation on the environment file after loading it.'''
+    possible_keys = {'imported_envs', 'channels', 'packages'}
```
bnemanich, line 127: Other possible keys: `git_tag_for_env`, `git_tag`

tests/test-env-invalid1.yaml
```diff
@@ -0,0 +1,20 @@
...
+# Licensed Materials - Property of IBM
+#
+# (C) Copyright IBM Corp. 2020. All Rights Reserved.
+#
+# US Government Users Restricted Rights - Use, duplication or
+# disclosure restricted by GSA ADP Schedule Contract with IBM Corp.
+#
+# *****************************************************************
+
+chnnels:
```
bnemanich, line 12: Add a comment here so that it is easy to see where the error should be.

tests/envlint.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python
```
bnemanich, line 1: Needs a copyright added.

Also, add a comment describing the purpose of this script.

bnemanich (approved):
Looks great!

PR #13, opened 2020-09-14
Add source style

bethune-bryant:
I added a pylint check for source style and updated our code to meet the default style (except line length of 125 instead of 100, which is the github line length).


bnemanich (approved):
This looks good.

.pylintrc
```diff
@@ -0,0 +1,588 @@
+[MASTER]
```
jayfurmanek, line 1: Is this just straight generated, or did you make any changes?

bethune-bryant: Almost straight generated default.
I increased the max line length from 100 to 125.

jayfurmanek: OK. How compatible is the rc file between version? Like if we start using a new pylint do we need to regenerate?

bethune-bryant: I'm not sure. But you raise a good point. We've seen pylint failures because of changes in versions in the past.
I'll make a change to pin to a specific pylint version. Maybe yamllint as well.

bethune-bryant: @jayfurmanek 
I pinned all of the test package versions.

jayfurmanek: Thanks. We'll likely want to revisit the rc when we update the pin. We'll see how it goes.

bethune-bryant: We may have to regenerate between versions, but that should be a conscious decision now.

jayfurmanek (approved):
Thanks!

Issue #7, opened 2020-09-10
Pytorch 1.6 build error


dustinvanstee:
Hello, I forked the repository and ran the following command to try and build a pytorch 1.6 release on a power8 minsky server ...

python build_env.py ../envs/pytorch-env.yaml

Here was the error I had currently.

ERROR: local_src_dir path "/gpfs/home/s4s004/vanstee/2020-09-opence/open-ce/builder/tensorrt-feedstock/config/../../local_files" specified doesn't exist
Unable to build recipe: tensorrt-feedstock

Also, once this process completes, how would I conda install the build into a virtual environment that I have ?  Does it make *whl files or something like that?

bnemanich:
Hi, 

The TensorRT repo requires a manual download of the TensorRT packages from Nvidia's website. For more details, please see: https://github.com/open-ce/tensorrt-feedstock. You should create a new directory called `local_files` within the directory where you will be running the build_env script from. The TensorRT tar file should go within this directory.

The output of the build_env script will be a conda channel (by default it will be called `condabuild`). You can install packages from this channel into your current conda environment by running `conda install -c ./condabuild PACKAGE_NAME`.

bnemanich:
Just added some new documentation with some of this information (#8). The main README.md will have those changes. Please follow those instructions and let us know how it goes.

jayfurmanek:
Closing this one. Please reopen if you find the documentation unclear.

dustinvanstee:
Hi thanks for the updates.  Here is an update : so I downloaded the tensorrt archive (both TensorRT-7.0.0.11.CentOS-7.6.ppc64le-gnu.cuda-10.2.cudnn7.6.tar.gz and TensorRT-7.2.0.13.CentOS-8.1.ppc64le-gnu.cuda-11.0.cudnn8.0.tar.gz).  I then created a directory called local_files in my build director and place the archives in the local_files directory.  

Next i ran this command
python build_env.py --python_versions 3.7  --conda_build_config ../conda_build_config.yaml     ../envs/opence-env.yaml

I still get this error 
```
Failure building recipe: tensorrt
Unable to build recipe: tensorrt-feedstock
```

My system is a power9 with Rhel 8, and CUDA 11.0.   Should i be modifying the defaults to run this on my system config below ?  

kernel : 4.18.0-147.20.1.el8_1.ppc64le
 
NVIDIA-SMI 450.36.06    Driver Version: 450.36.06    CUDA Version: 11.0

[...]

PR #18, opened 2020-09-15
Add functionality to build_env for building in docker containers

bnemanich:
This PR will add the `--docker_build` option to `build_env.py`. This option will tell the build_env script to build a docker image and run the actual build_env job within a container based on that image. The purpose of this feature is to allow the user to easily build packages within a clean environment.

### DIFF
--- a/builder/docker_build.py
+++ b/builder/docker_build.py
@@ -1,0 +1,69 @@
+"""
+*****************************************************************
+Licensed Materials - Property of IBM
+(C) Copyright IBM Corp. 2020. All Rights Reserved.
+US Government Users Restricted Rights - Use, duplication or
+disclosure restricted by GSA ADP Schedule Contract with IBM Corp.
+*****************************************************************
+"""
+import os
+
+OPEN_CE_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+BUILD_IMAGE_PATH = os.path.join(OPEN_CE_PATH, "images/builder")
+LOCAL_FILES_PATH = os.path.join(os.path.dirname(OPEN_CE_PATH), "local_files")
+HOME_PATH = "/home/builder"
+
+REPO_NAME = "open-ce"
+
+def build_image():
+    """
+    Build a docker image from the Dockerfile in BUILD_IMAGE_PATH.
+    Returns a result code and the name of the new image.
+    """
+    image_name = REPO_NAME + ":open-ce-builder-" + str(os.getuid())
+    build_cmd = "docker build "
+    build_cmd += "-f " + os.path.join(BUILD_IMAGE_PATH, "Dockerfile") + " "
+    build_cmd += "-t " + image_name + " "
+    build_cmd += "--build-arg BUILD_ID=" + str(os.getuid()) + " "
+    build_cmd += "--build-arg GROUP_ID=" + str(os.getgid()) + " "
+    build_cmd += "."
+
+    result = os.system(build_cmd)
+
+    return result, image_name
+
+def build_in_container(image_name, output_folder, arg_strings):
+    """
+    Run a build inside of a container using the provided image_name.
+    """
+    docker_cmd = "docker run --rm "
+
+    # Add output folder
+    local_output_folder = os.path.join(os.getcwd(), output_folder)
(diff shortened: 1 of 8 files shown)


builder/docker_build.py
```diff
@@ -0,0 +1,69 @@
...
+    docker_cmd = "docker run --rm "
+
+    # Add output folder
+    local_output_folder = os.path.join(os.getcwd(), output_folder)
+    if not os.path.isdir(local_output_folder):
+        os.mkdir(local_output_folder)
+    docker_cmd += "-v " + local_output_folder + ":" + os.path.join(HOME_PATH, output_folder) + ":rw "
+
+    # Add open-ce directory
+    docker_cmd += "-v " + OPEN_CE_PATH + ":" + HOME_PATH + "/open-ce:ro "
```
dllehr81, line 48: If possible, I'd still recommend copying the src into the Dockerfile if only to keep a sanitized version.  If someone were sharing this directory, you'd possibly end up with dirty code.  Plus if you're already doing the docker build as part of the process, you have the guarantee that it'll be up to date with the image.

builder/docker_build.py
```diff
@@ -0,0 +1,69 @@
...
+
+    # Add local_files directory (if it exists)
+    if os.path.isdir(os.path.join(os.getcwd(), "local_files")):
+        docker_cmd += "-v " + os.getcwd() + "/local_files:" + HOME_PATH + "/local_files:ro "
```
dllehr81, line 52: you use OPEN_CE PATH up above, but here you use `getcwd` is the idea that you would run this from a different directory?  Also you may have the same directories mounted twice, i.e. `local_files` may be under `HOME_PATH/local_files`  and `HOME_PATH/open-ce/local_files`

builder/docker_build.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # Add the open-ce directory
+    result = _copy_to_container(OPEN_CE_PATH, HOME_PATH, container_name)
+    if result:
+        print("Error copying open-ce directory into container")
+        return 1
+
+    # Add local_files directory (if it exists)
+    if os.path.isdir(os.path.join(os.getcwd(), "local_files")):
+        result = _copy_to_container(os.path.join(os.getcwd(), "local_files"), HOME_PATH, container_name)
```
bethune-bryant, line 97: Should you be using `LOCAL_FILES_PATH` here?

builder/docker_build.py
```diff
@@ -0,0 +1,133 @@
...
+BUILD_IMAGE_PATH = os.path.join(OPEN_CE_PATH, "images/builder")
+LOCAL_FILES_PATH = os.path.join(os.path.dirname(OPEN_CE_PATH), "local_files")
+HOME_PATH = "/home/builder"
+
+REPO_NAME = "open-ce"
+IMAGE_NAME = "open-ce-builder"
+
+DOCKER_TOOL = "docker"
+
+def build_image():
```
bethune-bryant, line 22: Does it make sense to have arguments for `REPO_NAME`, etc., and just have them default to the constants?
It could make it more flexible in the future.
If we do it here, we may want to do it for other functions in this file.

bethune-bryant (changes requested):

builder/docker_build.py
bnemanich: It might in the future, but I would like to wait until there is a use case for it.

builder/docker_build.py
```diff
@@ -11,7 +11,7 @@
 
 OPEN_CE_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
 BUILD_IMAGE_PATH = os.path.join(OPEN_CE_PATH, "images/builder")
-LOCAL_FILES_PATH = os.path.join(os.path.dirname(OPEN_CE_PATH), "local_files")
+LOCAL_FILES_PATH = os.path.join(os.path.join(os.getcwd(), "local_files")
```
bethune-bryant, line 14: Forgot a paren.

bethune-bryant (changes requested):

bethune-bryant (approved):
Looks good to me now.
Once this merges we may mention it over on #7

PR #20, opened 2020-09-17
add pytorch-extras-env yaml

cdeepali:
`pytorch-lightning` requires `tensorboard` at run time, so creating a `pytorch-extras-env.yaml` file. 

https://github.com/open-ce/open-ce/issues/37

### DIFF
--- a/envs/pytorch-lightning-env.yaml
+++ b/envs/pytorch-lightning-env.yaml
@@ -1,0 +1,6 @@
+imported_envs:
+  - pytorch-env.yaml
+  - tensorflow-env.yaml
+
+packages:
+  - feedstock : pytorch-lightning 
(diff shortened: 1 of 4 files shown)


envs/pytorch-lightning-env.yaml
```diff
@@ -0,0 +1,6 @@
+imported_envs:
+  - pytorch-env.yaml
+  - tensorflow-env.yaml
```
npanpaliya, line 3: I think if pytorch-lightning needs tensorboard, then we should also have a tensorboard env file, instead of using tensorflow env file. Tensorflow env file has got many other packages too like serving, hub, datasets, metadata which are coming soon.

cdeepali: Did I overlook, I didn't see a TB file yet.

cdeepali:
Added env file for TB and updated pytorch env file to import it for pytorch-lightning.

jayfurmanek:
Didn't we find that we could decouple the TB need? I'd rather not have seperate env files for TB and Pytorch-lightning.

jayfurmanek (approved):

PR #22, opened 2020-09-17
Add conda_builg_config validation.

### LINKED ISSUE #15: Validate `conda_build_config.yaml` as part of PR tests.
Meaning pre-verify there are no inherent conflict in what is listed - and to catch that before merge.
For example if we pin cffi to 0.14 to allow for newer python builds, but also pin another package that needs an older cffi (or libffi) then we'd have an immediate conflict. It would be good to catch that before merging

bethune-bryant:
Resolves #15

### DIFF
--- a/builder/build_env.py
+++ b/builder/build_env.py
@@ -237,15 +237,14 @@ def _create_recipes(repository, recipes, variant_config_files, variants, channel
         packages, run_deps, host_deps, build_deps, test_deps = _get_package_dependencies(recipe.get('path'),
                                                                                          variant_config_files,
                                                                                          variants)
-        output = { 'recipe' : recipe.get('name', None),
-                   'repository' : repository,
-                   'packages' : packages,
-                   'run_dependencies' : run_deps,
-                   'host_dependencies' : host_deps,
-                   'build_dependencies' : build_deps,
-                   'test_dependencies' : test_deps,
-                   'channels' : channels if channels else []}
-        outputs.append(output)
+        outputs.append({ 'recipe' : recipe.get('name', None),
+                         'repository' : repository,
+                         'packages' : packages,
+                         'run_dependencies' : run_deps,
+                         'host_dependencies' : host_deps,
+                         'build_dependencies' : build_deps,
+                         'test_dependencies' : test_deps,
+                         'channels' : channels if channels else []})
 
     os.chdir(saved_working_directory)
     return outputs
@@ -255,6 +254,9 @@ def create_all_recipes(env_config_files, variants, #pylint: disable=too-many-arg
                         git_location=DEFAULT_GIT_LOCATION,
                         git_tag_for_env="master",
                         conda_build_config=utils.DEFAULT_CONDA_BUILD_CONFIG):
+    '''
+    Create a recipe dictionary for each recipe needed for a given environment file.
+    '''
 
     result, env_config_data_list = load_env_config_files(env_config_files, variants)
     if result != 0:
--- a/builder/build_env.py
+++ b/builder/build_env.py
@@ -210,15 +210,17 @@ def _get_package_dependencies(path, variant_config_files, variants):
     # Parse out the package names and dependencies from each variant
     packages = set()
     deps = set()
+    versioned_deps = set()
     for meta,_,_ in metas:
         packages.add(meta.meta['package']['name'])
         deps.update(remove_versions(meta.meta['requirements'].get('run', [])))
+        versioned_deps.update(meta.meta['requirements'].get('run', []))
         deps.update(remove_versions(meta.meta['requirements'].get('host', [])))
         deps.update(remove_versions(meta.meta['requirements'].get('build', [])))
         if 'test' in meta.meta:
             deps.update(remove_versions(meta.meta['test'].get('requires', [])))
 
-    return packages, deps
+    return packages, deps, versioned_deps
 
 def _create_recipes(repository, recipes, variant_config_files, variants, channels):
     """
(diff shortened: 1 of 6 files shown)


builder/build_env.py
```diff
@@ -234,19 +236,66 @@ def _create_recipes(repository, recipes, variant_config_files, variants, channel
...
-                                                   variant_config_files,
-                                                   variants)
+        packages, deps, versioned_deps = _get_package_dependencies(recipe.get('path'),
+                                                                   variant_config_files,
+                                                                   variants)
         output = { 'recipe' : recipe.get('name', None),
                    'repository' : repository,
                    'packages' : packages,
                    'dependencies' : deps,
+                   'versioned_dependencies' : versioned_deps,
```
bethune-bryant, line 246: @bnemanich 
You can see here I added versioned dependencies to what is returned so that I could re-use these functions.

builder/build_env.py
```diff
@@ -234,19 +236,66 @@ def _create_recipes(repository, recipes, variant_config_files, variants, channel
...
                    'channels' : channels if channels else []}
         outputs.append(output)
```
bethune-bryant, line 248: @bnemanich
You can see here I added versioned dependencies to what is returned so that I could re-use these functions.

builder/build_env.py
```diff
@@ -234,19 +236,66 @@ def _create_recipes(repository, recipes, variant_config_files, variants, channel
...
 
     os.chdir(saved_working_directory)
     return outputs
 
+def _create_all_recipes(env_config_files, variants, #pylint: disable=too-many-arguments
```
bnemanich, line 253: This shouldn't start with an under score, since we will be using outside of this file.

builder/build_env.py
bnemanich: I think it would make sense to only have a field called `'dependencies'`. The versions from the dependencies can be removed on line 322 before they are used to create the dep tree. In the future, we might want to do something with the versions there anyways.

bethune-bryant: That makes sense. I'll make this change.

builder/build_env.py
```diff
@@ -234,19 +234,70 @@ def _create_recipes(repository, recipes, variant_config_files, variants, channel
...
+                                                                                         variant_config_files,
+                                                                                         variants)
+        outputs.append({ 'recipe' : recipe.get('name', None),
+                         'repository' : repository,
+                         'packages' : packages,
+                         'run_dependencies' : run_deps,
+                         'host_dependencies' : host_deps,
+                         'build_dependencies' : build_deps,
+                         'test_dependencies' : test_deps,
+                         'channels' : channels if channels else []})
```
bethune-bryant, line 247: @bnemanich 
I split all of the dependency types into their own set.

builder/build_env.py
```diff
@@ -270,7 +321,12 @@ def _create_dep_tree(recipes):
     outputs = []
     for index, recipe in enumerate(recipes):
         deps = []
-        for dep in recipe.get('dependencies', []):
+        dependencies = set()
+        dependencies.update({utils.remove_version(dep) for dep in recipe.get("run_dependencies")})
+        dependencies.update({utils.remove_version(dep) for dep in recipe.get("build_dependencies")})
+        dependencies.update({utils.remove_version(dep) for dep in recipe.get("host_dependencies")})
+        dependencies.update({utils.remove_version(dep) for dep in recipe.get("test_dependencies")})
+        for dep in dependencies:
```
bethune-bryant, line 329: @bnemanich 
And then I strip the versions and combine them all into a single set for the dependency tree.

builder/build_env.py
bethune-bryant: @bnemanich 
See comments on how I implemented this.

tests/configlint.py
```diff
@@ -0,0 +1,101 @@
...
+            print(str(env_file) + " : " + str(variant))
+            retval,recipes = build_env.create_all_recipes([env_file],
+                                                        variant,
+                                                        repository_folder=args.repository_folder,
+                                                        conda_build_config=args.conda_build_config)
+
+            packages = [package for recipe in recipes for package in recipe["packages"]]
+            deps = {utils.remove_version(dep) for recipe in recipes for dep in recipe.get("run_dependencies")}
+
+            cli = " -c https://public.dhe.ibm.com/ibmdl/export/pub/software/server/ibm-ai/conda/ "
```
bnemanich, line 88: The channels should be picked up from the env file, not hard coded here.

bethune-bryant: There are no channels in the `open-ce` env file.
Should it do a union of all the channels from all the imported env files?

bethune-bryant: Or should I add a channel to the open-ce-env file?

bnemanich: I think that is better. It is still not a perfect solution, since theoretically some of those environment files won't have that channel activated. But I think it is the better thing to do.

bethune-bryant:
@bnemanich 
Should we be pulling the variant possibilities from somewhere instead of hard coding them?

tests/configlint.py
bnemanich: By that, I mean a union of the channels.

bethune-bryant: Done.

bethune-bryant:
@bnemanich 
@jayfurmanek 
It should be good to go now.

bnemanich (approved):

jayfurmanek (approved):
Looks good now!

PR #24, opened 2020-09-18
add cuda images

jayfurmanek:
First stab at adding an image for CUDA and moving away from relying on the `cudatoolkit-dev` package from WML CE.

The main differences here vs the base (non-CUDA) image is the FROM line relying on Nvidia's base ubi7 image. And then manually removing the distro compilers.


bnemanich:
It feels like we shouldn't need different Dockerfiles for x86 and ppc.

Also, please see the changes that were made in #18 to the Dockerfile. There were a few changes to the GROUP_ID and USER_ID functionality that should be added here as well (along with a change in directories).

jayfurmanek:
yeah. Nvidia doesn't use multi-arch/manifest images. We could pass in some ARGS, but then would couldn't build with the docker on RH7 since its older.

bnemanich:
You should be able to do something like changing:
`FROM nvidia/cuda-ppc64le:${CUDA_VER}-devel-ubi7`
to
`FROM nvidia/cuda${ARCH}:${CUDA_VER}-devel-ubi7`

Since we already are using an environment variable for the CUDA version, adding an extra one for the ARCH shouldn't make things any more complicated.

jayfurmanek:
oops, I actually meant to not have that CUDA_VER there either. I had that, but it requires ARG, so I removed the ARG, left it in in the FROM line though.

@dllehr81 What do you think? If we use ARGS we're not able to build the image on RH7, right?

dllehr81:
So the RH7 part is true...however, podman which is available on RH7/RH8 fully supports ARG's on the first line. We'll just need to steer people in that direction if they don't have `docker-ce` 

For the location, I think we talked about in #18  that we'll try `images` on the root repository level to store these things in.  Probably save us an extra PR later on it.

bnemanich:
I think right now, the cost of maintaining multiple Dockerfiles is probably worth the benefit of people being able to use the standard version of `docker` on RHEL 7. In the future we can decide whether we can merge these files. Once we start adding more versions of CUDA, it could get to be kind of a pain.

I do think the images should be moved to the new `images` directory though.

jayfurmanek:
yeah, ok, I'll keep them separate for now. It won't be hard to merge them together using ARGs in the future.

bnemanich (approved):

dllehr81:
fine..ignore the future...I get it :)

PR #26, opened 2020-09-21
Loop create_all_recipes over each variant combination

bnemanich:
The main purpose of this PR is to only have a single dependency tree for an entire run of build_env. This will eventually make it easier to generate a YAML file for the dependency tree that can be used by an automated CI system. In a future PR, the code will be cleaned up a bit more by creating a dependency tree class.

This has a few behavioral changes. The build_type variants will now be split up into separate calls to build_feedstock, in the same manner as python variants currently are. This should allow for faster distributed jobs, since cpu and cuda versions of a package can now be built at the same time. This still relies on conda-build's `--skip-existing` argument.

The environment files have now been changed to use the `==` operator instead of `in`, since there will only ever be a single value for each variant when the environment file is being parsed. This will match up better with the way they are being used in the recipe's meta.yaml file.


bethune-bryant:
@bnemanich 
It looks like the build is failing because `nccl` and `tensorrt` aren't being found as dependencies, so they're trying to be installed.

bethune-bryant (changes requested):
I think the problem is on line 31:
https://github.com/open-ce/open-ce/blob/83bf311669d2730604086a54d0cd51988982f4c7/tests/configlint.py#L31

I'm making the variants as singleton lists, so `in` worked, but not `==`.

envs/pytorch-env.yaml
```diff
@@ -1,12 +1,12 @@
...
+{% if build_type == 'cuda' %}
 channels:
   - https://public.dhe.ibm.com/ibmdl/export/pub/software/server/ibm-ai/conda/
 {% endif %}
 
 packages:
   - feedstock : pytorch
   - feedstock : numactl
-{% if 'cuda' in build_type %}
+{% if build_type == 'cuda' %}
```
jayfurmanek, line 9: Hmm..while we are changing this, should we make the PyTorch and TF env files similar in how they mark CUDA only packages?
This file uses `if` statements while the TF ones marks each line indiividually.

bnemanich: OK, I changed it to using the if statements if there are more than 1 packages that need the conditional.

builder/build_env.py
```diff
@@ -362,28 +368,27 @@ def build_env(arg_strings=None):
...
-        print("Builds for python version: " + py_vers)
-        python_build_args = ["--python_versions", py_vers]
-        variants = { 'python' : py_vers, 'build_type' : utils.parse_arg_list(args.build_types) }
-
-        result, recipes = create_all_recipes(args.env_config_file, variants,
+    # Create a dependency tree that includes recipes for every combination
+    # of variants.
+    variants = { 'python' : utils.parse_arg_list(args.python_versions),
+                 'build_type' : utils.parse_arg_list(args.build_types) }
+    variant_cart_product = [dict(zip(variants,y)) for y in product(*variants.values())]
```
jayfurmanek, line 387: This is better than the big loop. :)

jayfurmanek (commented):
Test is still broke, but this is a better overall approach. Having a complete tree for the entire thing is helpful, especially for the CI.

builder/build_env.py
```diff
@@ -405,23 +410,27 @@ def build_env(arg_strings=None):
...
+        package_build_args = ["--working_directory", recipe['repository']]
 
-            for channel in recipe.get('channels', []):
-                package_build_args += ["--channels", channel]
+        for channel in recipe.get('channels', []):
+            package_build_args += ["--channels", channel]
 
-            if 'recipe' in recipe:
-                package_build_args += ["--recipes", recipe['recipe']]
+        package_build_args += ["--python_versions", recipe['python']]
```
bnemanich, line 423: @bethune-bryant I had to move these before adding the `--recipes` option. It looks line some of the tests rely on `--recipes` being the last argument passed to `build_feedstock`.

bethune-bryant: Yeah, it looks like `validate_build_feedstock` is grabbing the last argument as the package being built.

bethune-bryant (approved):
Looks good to me.

PR #31, opened 2020-09-22
Cleanup `configlint.py` into integration tool.

### LINKED ISSUE #30: Cleanup `configlint.py` into integration tool.
The functionality in `tests/configlint.py` may be useful to users that are trying to establish a set of packages/versions. We should cleanup `configlint.py` and turn it into an externalized tool.

bethune-bryant:
Resolves #30


validation/validate_env.py
```diff
@@ -7,9 +7,7 @@
 US Government Users Restricted Rights - Use, duplication or
 disclosure restricted by GSA ADP Schedule Contract with IBM Corp.
 *****************************************************************
-"""
 
-"""
 *******************************************************************************
 Script: envlint.py
```
bnemanich, line 12: These comments need to be modified for the new script name and arguments.

tests/validate_config_test.py
```diff
@@ -0,0 +1,254 @@
+# *****************************************************************
```
bnemanich, line 1: It seems like some of this is duplicated in the build_env tests. Should some of it be moved to a common file?

open-ce/docker_build.py
```diff
@@ -15,7 +15,7 @@
 HOME_PATH = "/home/builder"
 
 REPO_NAME = "open-ce"
-IMAGE_NAME = "open-ce-builder"
+IMAGE_NAME = "open-ce-open-ce"
```
bnemanich, line 18: Maybe this should still be builder?

bethune-bryant: Thanks, I tried to only replace the occurrences of builder that needed to be replaced, but I must have missed this one.

bnemanich (approved):

PR #38, opened 2020-09-24
Add BuildTree class

bnemanich:
This PR creates 2 new classes.

1. BuildCommand
    Holds all of the information needed to perform a single run of build_feedstock. Provides two helpful methods:
    feedstock_args: This will return a list of strings that can be passed to build_feedstock to perform a build.
    name: This will print out a name for the command, combining the recipe name with the python versions and build_type.

2. BuildTree:
    Container for a list of BuildCommands.
    The constructor will clone all of the feedstocks specified in the provided environment files and create a list of BuildCommands using the dependency information found in the recipe's meta.yaml files.
    Iterating over the BuildTree will ensure that all dependent packages appear before any packages relying on them.
    Indexing into the BuildCommand will correlate to the indices provided in a BuildCommand's build_command_dependencies field.


open-ce/build_env.py
```diff
@@ -35,20 +35,12 @@
...
-          " for a list of requirements.")
-    sys.exit(1)
 
 import build_feedstock
+from build_tree import BuildTree, DEFAULT_GIT_LOCATION
 import docker_build
 import utils
+from utils import OpenCEError
 
 DEFAULT_GIT_LOCATION = "https://github.com/open-ce"
```
bethune-bryant, line 45: It looks like you're importing `DEFAULT_GIT_LOCATION`, but also define it here.

bethune-bryant (changes requested):
Other than that small issue, it looks good to me.

open-ce/build_env.py
bnemanich: Thanks! Should be fixed now.

bethune-bryant (approved):

PR #41, opened 2020-09-25
Add validate_env test

bethune-bryant:
I left out the unit test for validate env in #31.


bnemanich (approved):

bethune-bryant:
I'm not sure what causes Codecov to add these comments, I've only seen them a couple times, but it's kinda nice.

PR #46, opened 2020-09-28
Enable feedstock value in an env config file to contain a url

bnemanich:
This PR will allow for the `feedstock` value in an environment config file to contain a url, instead of just providing a name of the feedstock. This will enable feedstocks to be hosted in locations besides `git_location`.

An environment file can now have values like this:

```
packages:
  - feedstock : https://github.com/open-ce/srsly-feedstock.git
  - feedstock : https://github.com/open-ce/thinc-feedstock
  - feedstock : git@github.com:open-ce/spacy-feedstock.git
  - feedstock : transformers
```

### DIFF
--- a/open-ce/build_tree.py
+++ b/open-ce/build_tree.py
@@ -252,16 +252,26 @@ class BuildTree():
                 if _make_hash(package) in packages_seen:
                     continue
 
+                # If the feedstock value starts with https: or git@, treat it as a url. Otheriwse
+                # combine with git_location and append "-feedstock.git"
+                if package['feedstock'].startswith("https:") or package['feedstock'].startswith("git@"):
+                    git_url = package['feedstock']
+                    if not git_url.endswith(".git"):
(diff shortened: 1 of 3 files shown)


open-ce/build_tree.py
```diff
@@ -252,16 +252,26 @@ def _create_all_recipes(self, variants):
                 if _make_hash(package) in packages_seen:
                     continue
 
+                # If the feedstock value starts with https: or git@, treat it as a url. Otheriwse
+                # combine with git_location and append "-feedstock.git"
+                if package['feedstock'].startswith("https:") or package['feedstock'].startswith("git@"):
```
bethune-bryant, line 257: Should this also check for `http:` to make it a little more generic? In case the repo isn't on github.

bnemanich: Good idea. I added that to the list.

bethune-bryant (approved):

Issue #63, opened 2020-09-29
Enable build_env to run without conda-build installed with docker_build option


bnemanich:
When the docker_build option is passed to build_env, it should not require conda_build to be present. The importing of modules that require conda-build should occur after the checking parsing of arguments and determining if it will be a docker_build.

bnemanich:
PR #65 addresses this issue.

PR #65, opened 2020-09-30
Check conda-build only when not using docker_build

npanpaliya:
## Description
Fixes issue https://github.com/open-ce/open-ce/issues/63.

As part of fix for above issue, I've also moved all DEFAULT variables to utils.py. I was thinking of adding constants.py for this. Please suggest if we should create something like constants.py and move all of them there.

### DIFF
--- a/open-ce/build_env.py
+++ b/open-ce/build_env.py
@@ -83,7 +83,13 @@ def build_env(arg_strings=None):
     if args.docker_build:
         return docker_build.build_with_docker(args.output_folder, sys.argv)
 
+    # Checking conda-build existence if --docker_build is not specified
     utils.check_if_conda_build_exists()
+       
+    # Here, importing BuildTree is intentionally done after checking
+    # existence of conda-build as BuildTree uses conda_build APIs.
+    from build_tree import BuildTree  # pylint: disable=import-outside-toplevel
+
     result = 0
 
     common_package_build_args = []
@@ -99,7 +105,6 @@ def build_env(arg_strings=None):
         os.mkdir(args.repository_folder)
 
     # Create the build tree
-    from build_tree import BuildTree  # pylint: disable=import-outside-toplevel
     try:
         build_tree = BuildTree(env_config_files=args.env_config_file,
                                python_versions=utils.parse_arg_list(args.python_versions),
--- a/open-ce/build_env.py
+++ b/open-ce/build_env.py
@@ -85,7 +85,7 @@ def build_env(arg_strings=None):
 
     # Checking conda-build existence if --docker_build is not specified
     utils.check_if_conda_build_exists()
-       
+
     # Here, importing BuildTree is intentionally done after checking
     # existence of conda-build as BuildTree uses conda_build APIs.
     from build_tree import BuildTree  # pylint: disable=import-outside-toplevel
--- a/open-ce/build_env.py
+++ b/open-ce/build_env.py
@@ -37,7 +37,6 @@ import os
 import sys
 
 import build_feedstock
-from build_tree import BuildTree, DEFAULT_GIT_LOCATION
 import docker_build
 import utils
 from utils import OpenCEError
@@ -55,7 +54,7 @@ def make_parser():
     parser.add_argument(
         '--git_location',
         type=str,
-        default=DEFAULT_GIT_LOCATION,
+        default=utils.DEFAULT_GIT_LOCATION,
         help='The default location to clone git repositories from.')
 
     parser.add_argument(
@@ -84,6 +83,7 @@ def build_env(arg_strings=None):
     if args.docker_build:
         return docker_build.build_with_docker(args.output_folder, sys.argv)
 
+    utils.check_if_conda_build_exists()
     result = 0
 
     common_package_build_args = []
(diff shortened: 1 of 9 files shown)


open-ce/build_env.py
```diff
@@ -99,6 +99,7 @@ def build_env(arg_strings=None):
         os.mkdir(args.repository_folder)
 
     # Create the build tree
+    from build_tree import BuildTree
```
npanpaliya, line 102: This import is intentionally moved after check_if_conda_build_exists because BuildTree imports conda_build and its submodules.

npanpaliya:
@bethune-bryant - pylint is complaining with an error "import outside toplevel" which I can't avoid as importing buildtree at the top imports conda_build.api too and in case conda-build isn't installed, it fails even before my check for conda-build "check_if_conda_build_exists".

bnemanich:
Looks good! Yeah, you will need to add the pylint exception in order for it to work. I think we expected something like that. We can leave the constants in util for now. If we start adding more of them, maybe we can create a new file.

I like the `check_if_conda_build_exists` function, but I think it needs to be called a few other places. We have two other tools that the user can launch: `validate_config` and `validate_env`. As it stands now, those two will fail without an error message if conda-build doesn't exist.

open-ce/build_env.py
bethune-bryant: Could we have this import right under the `check_if_conda_build_exists` call, with a comment saying that the order is important, that way it's clearer?

open-ce/build_env.py
```diff
@@ -99,6 +99,7 @@ def build_env(arg_strings=None):
         os.mkdir(args.repository_folder)
 
     # Create the build tree
+    from build_tree import BuildTree  # pylint: disable=import-outside-toplevel
```
bethune-bryant, line 102: Could we have this import right under the check_if_conda_build_exists call, with a comment saying that the order is important, that way it's clearer?

bethune-bryant (changes requested):
Can you add a test for this that ensures the docker_build option works without conda_build, so we know if it gets broken in the future.

You could mock  pkg_resources.get_distribution to return false and mock `docker_build.build_with_docker` to ensure that it gets called still.

open-ce/validate_config.py
```diff
@@ -87,4 +87,5 @@ def validate_config(arg_strings=None):
     return 0
 
 if __name__ == '__main__':
+    utils.check_if_conda_build_exists()
```
bnemanich, line 90: I think these need to happen before importing build_tree. One other option is to put these checks back in the files that use conda_build so that it is easier to keep track of (you can still use the function call at the top of the file).

bethune-bryant:
This looks great, thank you @npanpaliya.
I agree with @bnemanich, It's okay to have some exceptions to Pylint. Here we're valuing the user experience over enforcing a single instance of a specific code style, so I think that's okay.

I also made a couple suggestions above.

bethune-bryant:
> Can you add a test for this that ensures the docker_build option works without conda_build, so we know if it gets broken in the future.
> 
> You could mock pkg_resources.get_distribution to return false and mock `docker_build.build_with_docker` to ensure that it gets called still.

@npanpaliya 
Do you think it's possible to add a test like this?

jayfurmanek (commented):
This looks good to me. I agree it would be good to add a test to ensure we don’t regress here but I think an unintentional reordering is unlikely.

npanpaliya:
> > Can you add a test for this that ensures the docker_build option works without conda_build, so we know if it gets broken in the future.
> > You could mock pkg_resources.get_distribution to return false and mock `docker_build.build_with_docker` to ensure that it gets called still.
> 
> @npanpaliya
> Do you think it's possible to add a test like this?

I've already added the tests. Please check build_env_test.py. Is that the same thing that you were saying?

tests/build_env_test.py
```diff
@@ -167,4 +168,17 @@ def test_build_env_docker_build(mocker):
 
     mocker.patch('docker_build.build_with_docker', return_value=0)
 
+    mocker.patch('pkg_resources.get_distribution', return_value=None)
+
     assert build_env.build_env(arg_strings) == 0
+
+def test_build_env_if_no_conda_build(mocker):
```
bnemanich, line 175: I think @bethune-bryant was asking for another test similar to this that:
1. Still patched pkg_resources.get_distribution to throw the `pkg_resources.DistributionNotFound` exception.
2. Patched `docker_build.build_with_docker` to return 0.
3. Passed in the "--docker_build" option as one of the values to `build_env`.

bethune-bryant: @bnemanich 
That's exactly what I was thinking. So we can make sure that even if conda-build doesn't exist the script can still do a docker build.
We'll still want to keep the test that you have here now as well, because we want to make sure that case fails.

npanpaliya: I also have the test case which you have mentioned here https://github.com/open-ce/open-ce/pull/65/files/e850576c4c10f0eac7a8c07e487f494eb159b2f4#diff-0ac0b7c73e283818f092b25e37520c43R171. To cover this scenario, I've modified the existing test case and patched pkg_resources.get_distribution to return None (instead of throwing an exception because get_distribution returns None if the package is not found and then throw the exception). So, effectively the test case ensures that even if conda-build is not present the docker_build.build_with_docker succeeds.

bnemanich (approved):
Looks great!

bethune-bryant (approved):
Looks good. Thank you @npanpaliya

Issue #44, opened 2020-09-28
Add External Dependencies to Open-CE Env Files


bethune-bryant:
If we add external dependencies to the open-ce env file, the compatibility of those versions and all the internal dependencies can be verified.

This is a step towards facilitating the creation of custom environments with packages outside the scope of Open-CE.

My idea is to add another section to the yaml files. Something like:

```yaml
imported_envs:
  - pytorch-env.yaml
  - tensorflow-env.yaml

{% if build_type == 'cuda' %}
channels:
  - https://public.dhe.ibm.com/ibmdl/export/pub/software/server/ibm-ai/conda/
{% endif %}

packages:
  - feedstock : horovod
  - feedstock : openmpi

external-packages:
  - package1 1.2
  - package2 2.3
```

bnemanich:
This looks good.

Do we want to be able to specify which channel an individual external package is coming from? Or just allow conda to look for it in any of the specified channels?

npanpaliya:
These external packages are meant to be just installed as dependencies before the build. Right?

bethune-bryant:
@npanpaliya These are really only meant to be used for validation purposes. This will allow users to ensure that external dependencies that they rely on are compatible in the environment file. They will never be installed during the build process.
These will also be used in the generation of the env file in #45.

bethune-bryant:
@bnemanich 

> Do we want to be able to specify which channel an individual external package is coming from? Or just allow conda to look for it in any of the specified channels?

The explicit spec file can have packages specified by the full path, as below. So we may be able to do something like that:
```
# This file may be used to create an environment using:
# $ conda create --name <env> --file <this file>
# platform: osx-64
@EXPLICIT
https://repo.anaconda.com/pkgs/free/osx-64/mkl-11.3.3-0.tar.bz2
https://repo.anaconda.com/pkgs/free/osx-64/numpy-1.11.1-py35_0.tar.bz2
https://repo.anaconda.com/pkgs/free/osx-64/openssl-1.0.2h-1.tar.bz2
https://repo.anaconda.com/pkgs/free/osx-64/pip-8.1.2-py35_0.tar.bz2
```

PR #67, opened 2020-09-30
Improve CE File Validation

bethune-bryant:

## Description

When working on #44 I noticed that the CE file validation wasn't correctly checking everything. Specifically that the sub-dictionaries for packages were correct.

This PR adds a much more general yaml schema validation method and enumerates the CE file keys.

### DIFF
--- a/open-ce/build_tree.py
+++ b/open-ce/build_tree.py
@@ -253,14 +253,16 @@ class BuildTree():
 
                 # If the feedstock value starts with any of the SUPPORTED_GIT_PROTOCOLS, treat it as a url. Otherwise
                 # combine with git_location and append "-feedstock.git"
-                if any(package[env_config.ENV_CONFIG_KEYS[env_config.Key.FEEDSTOCK]].startswith(protocol) for protocol in SUPPORTED_GIT_PROTOCOLS):
-                    git_url = package[env_config.ENV_CONFIG_KEYS[env_config.Key.FEEDSTOCK]]
+                feedstock_value = package[env_config.ENV_CONFIG_KEYS[env_config.Key.FEEDSTOCK]]
+                if any(feedstock_value.startswith(protocol) for protocol in SUPPORTED_GIT_PROTOCOLS):
+                    git_url = feedstock_value
                     if not git_url.endswith(".git"):
                         git_url += ".git"
                     repository = os.path.splitext(os.path.basename(git_url))[0]
                 else:
-                    git_url = self._git_location + "/" + package[env_config.ENV_CONFIG_KEYS[env_config.Key.FEEDSTOCK]] + "-feedstock.git"
-                    repository = package[env_config.ENV_CONFIG_KEYS[env_config.Key.FEEDSTOCK]] + "-feedstock"
+                    git_url = "{}/{}-feedstock.git".format(self._git_location, feedstock_value)
+
+                    repository = feedstock_value + "-feedstock"
 
                 # Check if the directory for the feedstock already exists.
                 # If it doesn't attempt to clone the repository.
--- a/open-ce/build_tree.py
+++ b/open-ce/build_tree.py
@@ -243,7 +243,6 @@ class BuildTree():
             return result, []
         packages_seen = set()
         recipes = []
-        external_deps = []
         # Create recipe dictionaries for each repository in the environment file
         for env_config_data in env_config_data_list:
 
--- a/open-ce/build_tree.py
+++ b/open-ce/build_tree.py
@@ -243,12 +243,11 @@ class BuildTree():
             return result, []
         packages_seen = set()
         recipes = []
+        external_deps = []
         # Create recipe dictionaries for each repository in the environment file
         for env_config_data in env_config_data_list:
 
-            packages = env_config_data.get('packages', [])
-            if not packages:
-                packages = []
+            packages = env_config_data.get(env_config.ENV_CONFIG_KEYS[env_config.Key.PACKAGES], [])
             for package in packages:
                 if _make_hash(package) in packages_seen:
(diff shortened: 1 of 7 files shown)


bethune-bryant (commented):
I found a couple other places I need to use the new enum:
https://github.com/open-ce/open-ce/pull/67/files#diff-5d6ec8262a3cf027f25917da374a3102L257-L265

I'll fix them tomorrow and look for any others.

open-ce/env_config.py
```diff
@@ -17,18 +18,46 @@
...
+
+_PACKAGE_SCHEMA ={
+    ENV_CONFIG_KEYS[Key.FEEDSTOCK]: utils.make_schema_type(str, True)
+}
+
+_ENV_CONFIG_SCHEMA = {
+    ENV_CONFIG_KEYS[Key.IMPORTED_ENVS]: utils.make_schema_type([str]),
+    ENV_CONFIG_KEYS[Key.CHANNELS]: utils.make_schema_type([str]),
+    ENV_CONFIG_KEYS[Key.GIT_TAG_FOR_ENV]: utils.make_schema_type(str),
+    ENV_CONFIG_KEYS[Key.GIT_TAG]: utils.make_schema_type(str),
```
bnemanich, line 47: This should actually go under _PACKAGE_SCHEMA

open-ce/env_config.py
```diff
@@ -17,18 +18,46 @@
...
+    Key.IMPORTED_ENVS: "imported_envs",
+    Key.CHANNELS: "channels",
+    Key.PACKAGES: "packages",
+    Key.GIT_TAG_FOR_ENV: "git_tag_for_env",
+    Key.GIT_TAG: "git_tag",
+    Key.FEEDSTOCK: "feedstock"
+}
+
+_PACKAGE_SCHEMA ={
+    ENV_CONFIG_KEYS[Key.FEEDSTOCK]: utils.make_schema_type(str, True)
```
bnemanich, line 40: There should also be a 'recipes' field here that can be a list of strings.

bnemanich: Key.CHANNELS can be in both the _ENV_CONFIG_SCHEMA and the _PACKAGE_SCHEMA

open-ce/env_config.py
```diff
@@ -17,18 +18,46 @@
           " for a list of requirements.")
     sys.exit(1)
 
+class Key: #pylint: disable=no-init,too-few-public-methods
+    '''Enum for Env Config Keys'''
+    IMPORTED_ENVS = 0
```
bnemanich, line 23: Instead of using integers here, I think you can put the actual string. That would make using this much simpler because you wouldn't have to index into ENV_CONFIG_KEYS.

open-ce/utils.py
```diff
@@ -103,3 +103,33 @@ def parse_arg_list(arg_list):
...
+    '''Validate a single type instance against a schema type.'''
+    if isinstance(schema_type, dict):
+        validate_dict_schema(value, schema_type)
+    else:
+        if not isinstance(value, schema_type):
+            raise OpenCEError("{} is not of expected type {}".format(value, schema_type))
+
+def validate_dict_schema(dictionary, schema):
+    '''Recursively validate a dictionary's schema.'''
+    for k, schema_item in schema.items():
```
bnemanich, line 121: Is it possible to do `for k, (required, schema_type)`?

bethune-bryant: Ooo maybe. I'll try that.

open-ce/env_config.py
bethune-bryant: That would simplify the usage, but you could never do something like iterate over all the values of the enum.

However. It looks like python 3.4 added some type of enum. Since we're only in Python 3 land now I'll look into this:
https://docs.python.org/3/library/enum.html

open-ce/build_tree.py
```diff
@@ -246,23 +246,23 @@ def _create_all_recipes(self, variants):
         # Create recipe dictionaries for each repository in the environment file
         for env_config_data in env_config_data_list:
 
-            packages = env_config_data.get('packages', [])
-            if not packages:
```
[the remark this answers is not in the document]
bethune-bryant: Wouldn't that make the get default to the empty list?
I'll do an example and see what happens.

bnemanich (changes requested):
I added some comments about the different schemas, as well as some cleanup stuff.

open-ce/env_config.py
bnemanich: Just to be clear: `GIT_TAG` should go under _PACKAGE_SCHEMA.

open-ce/env_config.py
bethune-bryant: Right now all of the CE files only have `feedstock` under `packages`.

open-ce/build_tree.py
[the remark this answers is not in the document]
bethune-bryant: You're right. Having an empty packages broke things. It also broke the yaml schema validator, so I fixed that also.

open-ce/utils.py
bethune-bryant: IT worked, so I added this.

open-ce/env_config.py
bethune-bryant: I changed to the Py3 way of doing enums. Let me know what you think.

open-ce/env_config.py
bethune-bryant: Fixed.

open-ce/env_config.py
bethune-bryant: Done.

bnemanich (approved):

Issue #25, opened 2020-09-18
PyAV Feedstock 


richardtml:
Thanks for the repos! 

Could you please provide a feedstock for [PyAV](https://github.com/PyAV-Org/PyAV)? It's a PyTorch official [dependency](https://pytorch.org/docs/stable/torchvision/index.html#torchvision.set_video_backend) to read video, however, ppc64le is disabled in [conda-forge](https://github.com/conda-forge/av-feedstock) channel.

jayfurmanek:
Thanks for the request @richardtml !
We are currently looking at some of the obscure packages needed within the TF and PyTorch ecosystems for a complete environment, so thanks for the highlight on `PyAv`.

A couple observations:
 - `torchvision` needs to be build with ffmpeg support enabled
 - PyAv looks to be 100% python, so it should work well as a noarch pkg across arch's

ffmpeg is fairly complex and has a lot of function of various licenses and some of which is patent encumbered. Luckily, Anaconda, does offer ffmpeg . We'll do some testing on if `torchtext`+`PyAv` works with Anaconda's `ffmpeg` without issue and if so, we'll pull PyAv in.

richardtml:
Thank you @jayfurmanek , I will keep an eye on this issue to test it as as soon as you publish it.

cdeepali:
@richardtml,  `torchvision` with `av` backend enabled is now available for test.

richardtml:
Thank you @cdeepali, I'll test it asap.

cdeepali:
Sure, Thanks @richardtml.

bnemanich:
Closing this issue, since `av` was added. @richardtml if you run into any problems, please create a new issue.

PR #69, opened 2020-10-01
add av feedstock

cdeepali:
https://github.com/open-ce/open-ce/issues/25

### DIFF
--- a/conda_build_config.yaml
+++ b/conda_build_config.yaml
@@ -8,6 +8,8 @@ autoconf:
   - 2.69
 automake:
   - 1.16
+av:
+  - 8.0.2
 bazel:
   - 3.*
 boost:
--- a/conda_build_config.yaml
+++ b/conda_build_config.yaml
@@ -9,7 +9,7 @@ autoconf:
 automake:
   - 1.16
 av:
-  - 8.0.2
+  - 8.*
 bazel:
   - 3.*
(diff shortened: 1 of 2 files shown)


npanpaliya (approved):

dncliss (approved):

dncliss (approved):

jayfurmanek (approved):
Looks good. We can rerun the linter once we push in the av feedstock.

jayfurmanek (changes requested):
Actually upon further review, let's loosen the pyav pin. According to https://github.com/PyAV-Org/PyAV/blob/main/CHANGELOG.rst  pyAv uses semantic versioning, which says:

```
Given a version number MAJOR.MINOR.PATCH, increment the:

MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards compatible manner, and
PATCH version when you make backwards compatible bug fixes.
Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.
```

So as long as they stay true to that we can change this pin to `8.*`

bnemanich (approved):

cdeepali:
@jayfurmanek, I have updated the `{av}` pin as suggested.

jayfurmanek (approved):

PR #77, opened 2020-10-06
Adding global feedstock PR action

bethune-bryant:

## Description

Adds a Github action to be used by feedstock PRs.

You can see how it's used here:

https://github.com/open-ce/.github/blob/60f6e74bffa17c7480a2a18160e215c1ba7ac106/workflow-templates/feedstock-pr.yml#L18


jayfurmanek:
If I was looking to add another optional test, would I:
 - add another step to the `action.yml` that calls into an new .py file for each new test
 - or would I add additional test directly to the `feedstock-optional-pr.py` file?

If its the first case,  we should give the files/tests it calls into more specific names
If its the second, we should perhaps split the code that does the testing into their own functions.

dncliss:
Or.... maybe a check that if a new *.py or *.sh file is present, there is some evidence of a copyright added.  Something similar to a successful "head -10 | grep -iq copyright" on the file.

bethune-bryant:
> 
> 
> If I was looking to add another optional test, would I:
> 
>     * add another step to the `action.yml` that calls into an new .py file for each new test
> 
>     * or would I add additional test directly to the `feedstock-optional-pr.py` file?
> 
> 
> If its the first case, we should give the files/tests it calls into more specific names
> If its the second, we should perhaps split the code that does the testing into their own functions.
@jayfurmanek 
After giving it some thought, I think that the best way to add a new test would be to add another step to the `action.yml` file. This would let us leverage the workflow infrastructure and keep us from basically re-implementing a testing framework.

I agree that I should change the python file and test step to better reflect that this step and python file is specifically to test the build numbers.

I'll make this change tomorrow.

bethune-bryant:
> 
> 
> Or.... maybe a check that if a new *.py or *.sh file is present, there is some evidence of a copyright added. Something similar to a successful "head -10 | grep -iq copyright" on the file.

@dncliss 
This could be a good next test to add once we have the infrastructure up and running.

bethune-bryant:
@jayfurmanek @bnemanich @dncliss 
I made the changes we discussed, and this should be ready to go.

bnemanich (approved):

.github/actions/feedstock-optional-pr/action.yml
```diff
@@ -0,0 +1,18 @@
...
+  steps:
+    - name: Install dependencies
+      shell: bash -l {0}
+      run: |
+        conda install -y conda-build
+    - name: Run PR Tests
+      shell: bash -l {0}
+      run: |
+        test_path="$GITHUB_ACTION_PATH/../../../tests"
+        python $test_path/feedstock_tests/feedstock-optional-pr.py
```
jayfurmanek, line 18: I think this needs changed to `check_build_numbers.py` ?

bethune-bryant: @jayfurmanek 
You're right. Thanks for catching this. It's fixed now.

Issue #45, opened 2020-09-28
Produce Conda Env File as part of `build_env`


bethune-bryant:
After #44 is complete it would be good to have the build scripts produce a conda env file that can be used to install the conda environment that was just validated and built.

This should be produced separate from the actual build process, similar to the DAG production, so that there is flexibility to integrate into future tools.

The environment file should be displayed at the end of the `build_env` run so that users can use that to setup a new conda environment.

npanpaliya:
I just tried to write small prototype code using BuildTree class that would generate an env file and I'm able to create an env from it.

bethune-bryant:
@npanpaliya Thanks for picking this up!

You should be able to just use the run_dependencies member in the BuildTree class:
https://github.com/open-ce/open-ce/blob/cfbe5c9b5cd84bc747745681d77030473c605acb/open-ce/build_tree.py#L53

Also, something I left out of the original issue description is that we want to display the environment file at the end of the `build_env` run so that users can use that to setup their new conda environment.(I added it there also)

npanpaliya:
Thanks @bethune-bryant . Yes, I'll use `run_dependencies`. 
Also, thanks for clarifying my doubt which I was about to ask here about conda env file generation. My doubt was should it be optional and controlled via command line argument to build_env.py script OR it should always be generated.

npanpaliya:
Also, one more point to confirm - There will be multiple conda env file based on build_types (2) and python_versions (3). So, at max 2*3=6 conda env files that we will be writing.

npanpaliya:
@jayfurmanek , @bnemanich ,@bethune-bryant - 
While working on this, I've come across a couple of problems - 

1.  BuildTree specific
   a. In BuildTree, we haven't considered conda_build_config.yaml which is inside config directory while `_get_package_dependencies` so that the versions of packages like `_xgboost_select` or `_tensorflow_select` are read. I've fixed this.
  b. run_dependencies gives runtime dependencies of build time dependencies of a package. For e.g. bazel's build time dependency is java-openjdk whose runtime dependencies are all those nss*, etc. from cdts-feedstock. So, ideally those packages aren't required for bazel to run. I think this can be handled by adding some tag for master package in each environment file. And run time dependencies for that master package should be taken to create conda env file. More thoughts or suggestions are needed on this.

2. Inconsistency in our recipes while specifying package name and version in requirement section.
[...]

PR #90, opened 2020-10-12
Add mpi type

bethune-bryant:
## Checklist before submitting

- [x] Did you read the [contributor guide](https://github.com/open-ce/open-ce/blob/master/CONTRIBUTING.md)?
- [x] Did you update any affected [documentation](https://github.com/open-ce/open-ce/blob/master/doc/)?
- [x] Did you write any [tests](https://github.com/open-ce/open-ce/blob/master/tests/) to validate this change?  

## Description

Fixes #50

PR #94, opened 2020-10-14
build_env should produce conda env file

npanpaliya:

## Description

Fixes # https://github.com/open-ce/open-ce/issues/45.

## Review process to land 

### DIFF
--- a/open-ce/build_feedstock.py
+++ b/open-ce/build_feedstock.py
@@ -36,8 +36,6 @@ import yaml
 
 import utils
 
-DEFAULT_RECIPE_CONFIG_FILE = "config/build-config.yaml"
-
 def make_parser():
     ''' Parser input arguments '''
     arguments = [utils.Argument.CONDA_BUILD_CONFIG, utils.Argument.OUTPUT_FOLDER,
@@ -95,12 +93,12 @@ def load_package_config(config_file=None):
     file as an argument, it will be assumed that there is only one
     recipe to build, and it is in the directory called 'recipe'.
     '''
-    if not config_file and not os.path.exists(DEFAULT_RECIPE_CONFIG_FILE):
+    if not config_file and not os.path.exists(utils.DEFAULT_RECIPE_CONFIG_FILE):
         recipe_name = os.path.basename(os.getcwd())
         build_config_data = {'recipes':[{'name':recipe_name, 'path':'recipe'}]}
     else:
         if not config_file:
-            config_file = DEFAULT_RECIPE_CONFIG_FILE
+            config_file = utils.DEFAULT_RECIPE_CONFIG_FILE
         if not os.path.exists(config_file):
             print("Unable to open provided config file: " + config_file)
             return None, config_file
@@ -199,4 +197,5 @@ def build_feedstock(args_string=None):
     return result
 
 if __name__ == '__main__':
+    utils.check_if_conda_build_exists()
     sys.exit(build_feedstock())
--- a/open-ce/build_feedstock.py
+++ b/open-ce/build_feedstock.py
@@ -29,7 +29,6 @@ For usage description of arguments, this script supports use of --help:
 *******************************************************************************
 """
 
-import argparse
 import sys
 import os
 import yaml
@@ -40,16 +39,15 @@ def make_parser():
     ''' Parser input arguments '''
     arguments = [utils.Argument.CONDA_BUILD_CONFIG, utils.Argument.OUTPUT_FOLDER,
                  utils.Argument.CHANNELS, utils.Argument.PYTHON_VERSIONS,
-                 utils.Argument.BUILD_TYPES]
+                 utils.Argument.BUILD_TYPES, utils.Argument.MPI_TYPES]
     parser = utils.make_parser(arguments,
-                               description = 'Build conda packages as part of Open-CE',
-                               formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+                               description = 'Build conda packages as part of Open-CE')
 
     parser.add_argument(
         '--recipe-config-file',
         type=str,
         default=None,
-        help="""Path to the recipe configuration YAML file. The configuration
+        help="""R|Path to the recipe configuration YAML file. The configuration
 file lists paths to recipes to be built within a feedstock.
 
 Below is an example stating that there are two recipes to build,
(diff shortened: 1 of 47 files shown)


open-ce/conda_env_file_generator.py
```diff
@@ -0,0 +1,108 @@
...
+
+import re
+import os
+import yaml
+import utils
+
+class CondaEnvFileGenerator():
+    """
+    The CondaEnvData class holds all of the information needed to generate a conda
+    environment file while can be used to create a conda environment from the built packages.
```
bnemanich, line 18:
```suggestion
    environment file that can be used to create a conda environment from the built packages.
```

open-ce/conda_env_file_generator.py
```diff
@@ -0,0 +1,108 @@
...
+
+    #pylint: disable=too-many-instance-attributes,too-many-arguments
+    def __init__(self,
+                 python_versions=utils.DEFAULT_PYTHON_VERS,
+                 build_types=utils.DEFAULT_BUILD_TYPES,
+                 channels=None,
+                 output_folder=None,
+                 env_file_prefix=utils.CONDA_ENV_FILENAME_PREFIX,
+                 ):
+        self.python_versions = python_versions
```
bnemanich, line 29: It might be cleaner to add the parse_arg_list function here, instead of where self.python_versions is being used. (Same for build_types).

npanpaliya:
I'll push all the review comments fixes in one go.

open-ce/build_tree.py
```diff
@@ -93,6 +93,9 @@ def _create_recipes(repository, recipes, variant_config_files, variants, channel
     os.chdir(repository)
 
     config_data, _ = build_feedstock.load_package_config()
+    feedstock_conda_build_config_file = build_feedstock.get_conda_build_config()
+    if feedstock_conda_build_config_file:
+        variant_config_files.append(feedstock_conda_build_config_file)
```
bethune-bryant, line 98: Was the recipe config file not being used?
We probably don't want to modify `variant_config_files`, because this will change the argument where this function gets called.

npanpaliya: Yes, earlier it wasn't being used due to which the contents in those files weren't used while creating build_tree. In my case,versions for select packages of xgboost, tf or pytorch weren't being populated. So, this fixes that problem.

open-ce/build_tree.py
```diff
@@ -331,6 +334,8 @@ def _traverse_deps_tree(self, is_seen, deps):
         for dep in deps:
             if is_seen[dep]:
                 continue
+            if self.build_commands[dep].build_command_dependencies is None:
+                self.build_commands[dep].build_command_dependencies = []
```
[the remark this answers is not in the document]
npanpaliya: Yes.

open-ce/build_tree.py
bethune-bryant, line 98: Okay, the recipe_config_file is being returned by `build_feedstock.load_package_config()` on line 95, but is being ignored.

I think this should be something more like:
```suggestion
    config_data, feedstock_conda_build_config_file  = build_feedstock.load_package_config()
    if feedstock_conda_build_config_file:
        variant_config_files.append(feedstock_conda_build_config_file)
```

But I'm still not sure if we should be modifying `variant_config_files`

open-ce/conda_env_file_generator.py
```diff
@@ -0,0 +1,108 @@
...
+            for dep in dependencies:
+                self.dependency_dict[key].add(_cleanup_depstring(dep))
+
+    def update_conda_env_file_content(self, build_command, build_tree):
+        """
+        This function updates dependency dictionary for each build command with
+        its dependencies both internal and external.
+        """
+
+        key = "py" + build_command.python + "-" + build_command.build_type
```
bnemanich, line 87: This key is getting generated in several places. It will get harder to maintain as we add more variants. Maybe it could be a method to build command (something like `build_command.variant_key()`). Then we would only have to update it in one place when more variants are added (see #90 )

open-ce/build_tree.py
npanpaliya: recipe_config_file returned by `build_feedstock.load_package_config()` is different than the one I want. I want `config/conda_build_config.yaml` whereas the config file returned by `load_package_config()` is `config/build-config.yaml`.

open-ce/build_feedstock.py
```diff
@@ -87,6 +87,14 @@ def make_parser():
 
     return parser
 
+def get_conda_build_config():
+    '''
+    Checks for a conda_build_config file inside config dir of the feedstock.
+    And returns the same if it exists.
+    '''
+    recipe_conda_build_config = os.path.join(os.getcwd(), "config", "conda_build_config.yaml")
```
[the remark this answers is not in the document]
npanpaliya: `utils.DEFAULT_RECIPE_CONFIG_FILE` is `"config/build-config.yaml"` in utils.py. In my logic, I want `config/conda_build_config.yaml`.

open-ce/conda_env_file_generator.py
npanpaliya: Yes, I'll create a method in BuildCommand for this. One thing I wanted to know is if BuildCommand will ever have python and build_type as None.

npanpaliya: I've instead created the method in utils.py as this variant key was needed irrespective of BuildCommand.

bethune-bryant:
Here is the generalize version script, if you usde it here, we'll probably want to move it up to utils.py.

https://github.com/open-ce/open-ce/blob/634d9bfd7aa134973f57640f35420bd95a931212/open-ce/validate_config.py#L29-L36

npanpaliya:
> Here is the generalize version script, if you use it here, we'll probably want to move it up to utils.py.
> 
> https://github.com/open-ce/open-ce/blob/634d9bfd7aa134973f57640f35420bd95a931212/open-ce/validate_config.py#L29-L36

Yeah, thanks @bethune-bryant . I'll use it along with my cleanup_depstring.

I will also update the PR with adding `build_string` to the dependency list to be more explicit. That's how conda env export dumps the dependencies.

open-ce/build_tree.py
bethune-bryant: This makes sense now. 
I don't think we should be modifying the input argument, as this will modify it in the calling function as well.

open-ce/conda_env_file_generator.py
```diff
@@ -0,0 +1,93 @@
...
+        self.dependency_dict = {}
+        self.channels = []
+        self._initialize_dependency_dict()
+        self._initialize_channels(channels, output_folder)
+
+    def _initialize_dependency_dict(self):
+        for py_version in self.python_versions:
+            for build_type in self.build_types:
+                key = utils.variant_key(py_version, build_type)
+                self.dependency_dict[key] = set()
```
bethune-bryant, line 40: This is missing the MPI_TYPE variant. And isn't very scalable.
Instead of a nested for loop here for all variant types, I think you may want to iterate through the possible variants with this function:
https://github.com/open-ce/open-ce/blob/035395fec283c51e7489f18a82eff678e379335c/open-ce/utils.py#L120-L125

open-ce/utils.py
```diff
@@ -150,3 +152,39 @@ def get_output(command):
...
+    return result
+
+def generalize_version(package):
+    """Add `.*` to package versions when it is needed."""
+
+    # Remove multiple spaces or tabs
+    package = re.sub(' +', ' ', package)
+
+    # Check if we want to add .* to the end of versions
+    py_matched = re.match(r'(\w+[-]*\w+)([\s,=,<,>]*)(.*)', package)
```
bethune-bryant, line 174: These may be better regexes to use.
```suggestion
    package = re.sub('\s+', ' ', package)

    # Check if we want to add .* to the end of versions
    py_matched = re.match(r'([\w-]+)([\s=<>]*)(.*)', package)
```

open-ce/utils.py
```diff
@@ -150,3 +152,39 @@ def get_output(command):
...
+    if py_matched:
+        name = py_matched.group(1)
+        operator = py_matched.group(2)
+        version = py_matched.group(3)
+        append_asterik = False
+
+        if len(version) > 0 and len(operator) > 0:
+
+            #Append .* at the end if it is not there and if operator is space or == or = or empty
+            if not version.endswith(".*") and operator.strip() in ["=", "==", " ", ""]:
```
bethune-bryant, line 185: If you're doing `operator.strip()`, I don't think you need both `" "` and `""`.
Also, I don't think we need to add th `.*` with just a single `=`.

npanpaliya: "" was needed when operator contains all the spaces. In that case strip() removes all the spaces returns empty string. 
For `=`, yes we can remove that. However, I'd kept it for consistency since we are adding .* for ==, although for =, .* isn't needed.

npanpaliya:
Thanks @bethune-bryant and @bnemanich  for your review. I'm implemented the suggested changes. 

I've not added build_string changes as it won't be possible to retrieve for any dependency except the main packages in the buildcommand. For e.g. if tensorflow-base is being built, we can get build_string for it from its meta.yaml. But if the same tensorflow-base is added a dependency for some other package like tensorflow-text, then it would just be a string and not the recipe. So, for that I won't get build_string. Then the dependencies I'll get in the final list will have two entries for tensorflow-base - one with build string and another without. So, I think it is better to avoid it. With current code in my PR, I'm able to create pytorch env. I'll re-verify TF, xgboost, horovod with the latest changes.

open-ce/conda_env_file_generator.py
npanpaliya: Yes, I've incorporated MPI_TYPE. I hadn't noticed that this PR got merged and I didn't pull it.

npanpaliya:
@bnemanich , @bethune-bryant , @jayfurmanek - Please let me know if you have further comments on this.

open-ce/build_tree.py
bnemanich: I agree with @bethune-bryant , we shouldn't be modifying the input argument. We might not be using variant_config_files right now after calling _create_recipes, but if we do later on it might lead to some confusing behavior. We can just create a new local variable for this change to keep things cleaner.

bnemanich:
I made one more comment. Other than that, I think it is good to go. (There are some linting errors that need to be fixed as well for the tests to pass).

open-ce/build_tree.py
npanpaliya: Okay. This is done.

bnemanich (approved):

PR #101, opened 2020-10-15
Adding release scripts

### LINKED ISSUE #100: Create script to perform tagging across all repos.
We need a script to automatically create tags across all repos for a release.

Once a decision is reached in #49

bethune-bryant:

## Description

Fixes #100

### DIFF
--- a/git_tools/create_opence_release.py
+++ b/git_tools/create_opence_release.py
@@ -82,6 +82,8 @@ def _main(arg_strings=None):
     print("--->Updating env files.")
     update_env_files(primary_repo_path, version_name)
 
+    git_utils.commit_changes(primary_repo_path, "Updates for {}".format(version_name))
+
     return
 
     tag_all_repos.tag_all_repos(github_org=args.github_org,
@@ -94,7 +96,17 @@ def _main(arg_strings=None):
 
 def update_env_files(open_ce_path, new_git_tag):
     for env_file in glob.glob(os.path.join(open_ce_path, "envs", "*.yaml")):
-        print(env_file)
+        print("--->Updating {}".format(env_file))
+        with open(env_file, 'r') as content_file:
+            env_file_contents = content_file.read()
+        if not "git_tag_for_env" in env_file_contents:
+            env_file_contents = """{}
+git_tag_for_env:
+  - {}
+""".format(env_file_contents, new_git_tag)
+
+        with open(env_file, 'w') as content_file:
+            content_file.write(env_file_contents)
 
 if __name__ == '__main__':
     try:
--- a/git_tools/create_opence_release.py
+++ b/git_tools/create_opence_release.py
@@ -18,6 +18,7 @@ import sys
 import os
 import glob
 import git_utils
+import tag_all_repos
 
 def _make_parser():
     ''' Parser input arguments '''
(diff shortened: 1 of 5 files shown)


git_tools/tag_all_repos.py
```diff
@@ -0,0 +1,133 @@
...
+
+def _make_parser():
+    ''' Parser input arguments '''
+    parser = argparse.ArgumentParser(
+        description = 'Tag all repos in an organization.',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+    parser.add_argument(
+        'github_org',
+        type=str,
```
bnemanich, line 35: Should we have a default value here?

git_tools/tag_all_repos.py
```diff
@@ -0,0 +1,133 @@
...
+*****************************************************************
+Licensed Materials - Property of IBM
+(C) Copyright IBM Corp. 2020. All Rights Reserved.
+US Government Users Restricted Rights - Use, duplication or
+disclosure restricted by GSA ADP Schedule Contract with IBM Corp.
+*****************************************************************
+
+*******************************************************************************
+Script: tag_all_repos.py
+A script that can be used to create the same annotated git tag in all
```
bnemanich, line 12: Can you add a short example of using this? Maybe the bare minimum of what you would need to pass to the script? It could either be here, or in a README file somewhere.

git_tools/tag_all_repos.py
bethune-bryant: I kinda like having to be explicit with it, so there's less chance of somebody making random tags across an org.

git_tools/tag_all_repos.py
bethune-bryant: I added the example here. What do you think?

bethune-bryant: Yeah, it's still sorta in flux.

git_tools/create_opence_release.py
```diff
@@ -0,0 +1,114 @@
...
+    git_utils.create_tag(primary_repo_path, version_name, version_msg)
+
+    push = git_utils.ask_for_input("Would you like to push changes to primary repo?")
+    if push.startswith("y"):
+        print("--->Pushing branch.")
+        git_utils.push_branch(primary_repo_path, branch_name)
+        print("--->Pushing tag.")
+        git_utils.push_branch(primary_repo_path, version_name)
+
+    tag_all_repos.tag_all_repos(github_org=args.github_org,
```
bnemanich, line 86: Should this still happen, even if the user didn't enter "y"?

git_tools/create_opence_release.py
```diff
@@ -0,0 +1,114 @@
...
+                                branch=args.branch,
+                                repo_dir=args.repo_dir,
+                                pat=args.pat,
+                                skipped_repos=args.primary_repo)
+
+def _update_env_files(open_ce_path, new_git_tag):
+    for env_file in glob.glob(os.path.join(open_ce_path, "envs", "*.yaml")):
+        print("--->Updating {}".format(env_file))
+        with open(env_file, 'r') as content_file:
+            env_file_contents = content_file.read()
```
bnemanich, line 98: Should this be loaded with pyyaml's safe_load function?

git_tools/create_opence_release.py
bethune-bryant: Yeah, that lets a user basically do a dry-run of everything, by saying no here, and then later saying no to pushing the tags.

git_tools/create_opence_release.py
bethune-bryant: That wouldn't work with environment file that have jinja would it?
I don't actually want to parse the yaml, I just want to treat it as text.

bnemanich: Ah, that is a good point.

bnemanich (approved):

jayfurmanek (approved):
This looks good! We're accumulating a good set of utils!

PR #117, opened 2020-10-21
Use new features from #94 in check_build_numbers

### LINKED ISSUE #106: Improve build_number check with new features from #94
Once #94 is merged I would like to clean [check_build_numbers.py](https://github.com/open-ce/open-ce/blob/master/tests/feedstock_tests/check_build_numbers.py) up to use some features that are being added. As well as adding a loop to loop through all variants and check build numbers for each variant combination.

bethune-bryant:

## Description

Fixes #106

### DIFF
--- a/tests/feedstock_tests/check_build_numbers.py
+++ b/tests/feedstock_tests/check_build_numbers.py
@@ -46,52 +46,56 @@ def main(arg_strings=None):
 
     config = get_or_merge_config(None)
     config.variant_config_files = [utils.DEFAULT_CONDA_BUILD_CONFIG]
-    recipe_conda_build_config = os.path.join(os.getcwd(), "config", "conda_build_config.yaml")
-    if os.path.exists(recipe_conda_build_config):
-        config.variant_config_files += [recipe_conda_build_config]
+
+    recipe_conda_build_config = build_feedstock.get_conda_build_config()
+    if recipe_conda_build_config:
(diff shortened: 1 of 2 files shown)


tests/feedstock_tests/check_build_numbers.py
```diff
@@ -46,52 +46,56 @@ def main(arg_strings=None):
...
+        for recipe in build_config_data["recipes"]:
+            metas = conda_build.api.render(recipe['path'],
+                                        config=config,
+                                        variants=variant,
+                                        bypass_env_check=True,
+                                        finalize=False)
+            for meta,_,_ in metas:
+                master_build_numbers[meta.meta['package']['name']] = {"version" : meta.meta['package']['version'],
+                                                                    "number" : meta.meta['build']['number']}
+
```
npanpaliya, line 67: Do you think we should create a function for this section as it is doing same thing for master branch as well as PR branch? I know this is not called from anywhere else. But I thought  having a function will just ensure we are retrieving same information from both branches, in case something changes in future.

npanpaliya: Rest looks good to me.

bethune-bryant: Thanks, that's a good idea. I made this change.

bnemanich (approved):

PR #121, opened 2020-10-22
Replacing return codes with exceptions

### LINKED ISSUE #52: Replace return codes with exceptions
Many of the functions in the open-ce directory return a potential error code. This should be modified to having the functions throw exceptions when an error occurs.

The entry point scripts should catch any of the OpenCE exceptions and return an error code if there is a failure.

bethune-bryant:

## Description

Fixes #52

## Review process to land 

### DIFF
--- a/open-ce/docker_build.py
+++ b/open-ce/docker_build.py
@@ -11,6 +11,7 @@ import datetime
 import platform
 
 import utils
+from utils import OpenCEError
 
 OPEN_CE_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
 BUILD_IMAGE_NAME = "builder-cuda-" + platform.machine()
@@ -30,7 +31,7 @@ def make_parser():
 
     return parser
 
-def build_image():
+def _build_image():
     """
     Build a docker image from the Dockerfile in BUILD_IMAGE_PATH.
     Returns a result code and the name of the new image.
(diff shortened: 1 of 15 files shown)


npanpaliya:
@bethune-bryant - While you're doing this cleanup,  I also have a thought of having constants.py where we can put all the DEFAULT* variables which are right now in utils.py and some are in docker_build.py.

bethune-bryant:
> @bethune-bryant - While you're doing this cleanup, I also have a thought of having constants.py where we can put all the DEFAULT* variables which are right now in utils.py and some are in docker_build.py.

@npanpaliya 
This is a really good idea. I'll probably do this in a another issue to keep things clean.

open-ce/env_config.py
```diff
@@ -49,14 +49,13 @@ def _validate_config_file(env_file, variants):
...
-                            "import another environment file.")
+            raise OpenCEError("Content Error!:\n"
+                              "An environment file needs to specify packages or "
+                              "import another environment file.")
         utils.validate_dict_schema(meta_obj.meta, _ENV_CONFIG_SCHEMA)
         return meta_obj
     except (Exception, SystemExit) as exc: #pylint: disable=broad-except
-        print('***** Error in %s:\n  %s' % (env_file, exc), file=sys.stderr)
-        return None
+        raise OpenCEError('***** Error in {}:\n  {}'.format(env_file, str(exc))) from exc
```
npanpaliya, line 58: Just a nitpick - Adding asteriks to the error message helps it to be highlighted. So, probably it is good to have for all error messages.

cdeepali: Yeah, it is very easy to differentiate Open-CE errors from others when these special delimiters.

tests/build_env_test.py
```diff
@@ -190,9 +192,9 @@ def test_env_validate(mocker, capsys):
         side_effect=buildTracker.validate_build_feedstock
     )
     env_file = os.path.join(test_dir, 'test-env-invalid1.yaml')
-    assert build_env.build_env([env_file]) == 1
-    captured = capsys.readouterr()
-    assert "Unexpected key chnnels was found in " in captured.err
+    with pytest.raises(OpenCEError) as exc:
+        build_env.build_env([env_file])
+    assert "Unexpected key chnnels was found in " in str(exc.value)
```
cdeepali, line 197: typo `chnnels was`

cdeepali: Please ignore the comment, as @npanpaliya pointed out that this is to test the incorrectness itself.

open-ce/docker_build.py
```diff
@@ -112,26 +113,22 @@ def build_in_container(image_name, output_folder, arg_strings):
 
     result = _create_container(container_name, image_name, output_folder)
     if result:
-        print("Error creating docker container: " + container_name)
-        return result
+        raise OpenCEError("Error creating docker container: \"{}\"".format(container_name))
```
bnemanich, line 116: Do you think the smaller functions such as `_create_container` should be throwing the exception? I think the code would be a little cleaner, but maybe the error messages wouldn't be as specific.

bethune-bryant: I though it made sense for the internal functions to return error codes, since they're basically just simple wrappers for a system call. Let me take another look at them and see how much it cleans it up to switch those as well.

bethune-bryant: This is done, it is a lot cleaner this way.

open-ce/env_config.py
bethune-bryant: All errors will now start with `[OPEN-CE-ERROR-#]` as defined here: https://github.com/open-ce/open-ce/blob/76aed751074e17f6d1958758b31a2a7fa53408f6/open-ce/errors.py#L37

bnemanich (approved):

PR #132, opened 2020-10-29
Add dependency check to feedstock PR

### LINKED ISSUE #99: Test package dependencies as part of a feedstock PR
The validate_config script should be able to be used to validate that the dependencies in a feedstock haven't been messed up.

bethune-bryant:

## Description

Fixes #99 

You can see it running under required checks here:
https://github.com/open-ce/horovod-feedstock/pull/9

### DIFF
--- a/.github/actions/feedstock-required-pr/action.yml
+++ b/.github/actions/feedstock-required-pr/action.yml
@@ -17,7 +17,7 @@ runs:
         # Validate all env files with the new feedstock.
         open_ce_path="$GITHUB_ACTION_PATH/../../../open-ce"
         env_path="$GITHUB_ACTION_PATH/../../../envs"
-        python ${open_ce_path}/validate_config.py ${env_path}/*-env.yaml
+        python ${open_ce_path}/validate_config.py ${env_path}/*-env.yaml \
                                                   --conda_build_config \
                                                   ./conda_build_config.yaml \
                                                   --build_types cuda,cpu \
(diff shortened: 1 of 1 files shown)


.github/actions/feedstock-required-pr/action.yml
```diff
@@ -7,6 +7,19 @@ description: "Required global PR tests for Open-CE feedstocks."
...
+        conda install -y conda-build
+    - name: Check Dependencies
+      shell: bash -l {0}
+      run: |
+        # Validate all env files that use the current feedstock.
+        open_ce_path="$GITHUB_ACTION_PATH/../../../open-ce"
+        env_path="$GITHUB_ACTION_PATH/../../../envs"
+        feedstock=$(echo ${PWD##*/} | sed 's|\(.*\)-.*|\1|')
+        for f in ${env_path}/*.yaml; do
+          if grep -Fq "${feedstock}" ${f}; then
```
bnemanich, line 22: I don't think this will handle nested environment files properly. Maybe we just want to do this on the open-ce.yaml file if we are worried about build time.

bethune-bryant: I was more wanting to make sure that it ran on everything that used it. In case it's a feedstock outside of open-ce.
I could switch it to test all env files. I don't think we're really worried about time.

bethune-bryant: @bnemanich 
Okay, it's checking all env files now for the default variants and cuda+cpu.

bnemanich (approved):

PR #134, opened 2020-10-30
Intial commit of test_feedstock script

### LINKED ISSUE #114: Create test_feedstock script
Create a test_feedstock script. This script will take a conda environment file and path to a conda channel. Then it will:

1. Look for a tests directory. If it finds one, it will look for a `tests.yaml` file.
1. Load the YAML file. The YAML file will have a collection of bash commands.
1. Create a conda environment based off the provided conda environment file and channel.
1. Run each bash command. If any fail, print out the ones that failed and return an error code.
1. Remove the conda environment.

bnemanich:

## Description

Fixes #114.

This adds the initial version of the test_feedstock script. This script will read an Open-CE test file for a feedstock and run all of the commands within that file. All of the tests will be run within a conda environment created from the provided conda environment file.

In order to run tests for a repository:
1. `git clone https://github.com/open-ce/open-ce.git`
1. `git clone https://github.com/open-ce/MY_REPO.git`
1. `cd MY_REPO`
1. `../open-ce/test_feedstock.py --conda_env_file MY_CONDA_ENV_FILE`

For an example of the test file, please see: https://github.com/open-ce/xgboost-feedstock/pull/10

Some future work will include:
* Adding  a script that will run all of the tests for a given Open-CE environment file.
* Adding some sort of label to the tests that will signify whether they are short or long tests (or anything else).
* Specifying a directory that tests should be run from.
* Adding documentation for how to run tests and what a test file should look like.
* Verify that test files are valid.

## Review process to land 

### DIFF
--- a/open-ce/test_feedstock.py
+++ b/open-ce/test_feedstock.py
@@ -1,0 +1,223 @@
+#!/usr/bin/env python
+"""
+*****************************************************************
+Licensed Materials - Property of IBM
+(C) Copyright IBM Corp. 2020. All Rights Reserved.
+US Government Users Restricted Rights - Use, duplication or
+disclosure restricted by GSA ADP Schedule Contract with IBM Corp.
+*****************************************************************
+"""
+
+import datetime
+import os
+import tempfile
+import sys
+import subprocess
+
+import yaml
+
+import utils
+from utils import OpenCEError
+
+class TestCommand():
+    """
+    Contains a test to run within a given conda environment.
+
+    Args:
+        name (str): The name describing the test.
+        conda_env (str): The name of the conda environment that the test will be run in.
+        bash_command (str): The bash command to run.
+        create_env (bool): Whether this is the command to create a new conda environment.
+        clean_env (bool): Whether this is the command to remove a conda environment.
+    """
+    #pylint: disable=too-many-arguments
+    def __init__(self, name, conda_env=None, bash_command="", create_env=False, clean_env=False):
+        self.bash_command = bash_command
+        self.conda_env = conda_env
+        self.name = name
+        self.create_env = create_env
+        self.clean_env = clean_env
+
+    def get_test_command(self, conda_env_file=None):
+        """"
+        Returns a string of the test command.
+
+        Args:
+            conda_env_file (str): The name of the original conda environment file.
+                                  This is only needed when create_env is True.
+        """
+        output = ""
+        if self.create_env:
+            output += "conda env create -f " + conda_env_file + " -n " + self.conda_env + "\n"
+            return output
+
+        if self.clean_env:
+            output += "conda env remove -y -n " + self.conda_env + "\n"
+            return output
+
+        output += "CONDA_BIN=$(dirname $(which conda))\n"
+        output += "source ${CONDA_BIN}/../etc/profile.d/conda.sh\n"
+        output += "conda activate " + self.conda_env + "\n"
+        output += self.bash_command + "\n"
+
+        return output
+
+    def run(self, conda_env_file):
+        """
+        Runs the test.
+
+        Creates a temporary bash file, and writes the contents to `get_test_command` into it.
+        Runs the generated bash file.
+        Removes the temporary bash file.
+
+        Args:
+            conda_env_file (str): The name of the original conda environment file.
+        """
+        print("Running: " + self.name)
+        # Create file containing bash commands
+        with tempfile.NamedTemporaryFile(mode='w+t', dir=os.getcwd(), delete=False) as temp:
+            temp.write("set -e\n")
+            temp.write(self.get_test_command(conda_env_file))
+            temp_file_name = temp.name
+
+        # Execute file
+        test_process = subprocess.run(["bash", temp_file_name], stdout=subprocess.PIPE, check=False,
+                                      stderr=subprocess.STDOUT, universal_newlines=True)
+
+        # Remove file containing bash commands
+        os.remove(temp_file_name)
+
+        result = TestResult(self.name, test_process.returncode, test_process.stdout)
+
+        if test_process.returncode != 0:
+            result.display_failed()
+
+        return result
+
+class TestResult():
+    """
+    Contains the results of running a test.
+
+    Args:
+        name (str): The name of the test that was run.
+        returncode (int): The return code of the test that was run.
+        output (str): The resuling output from running the test.
+    """
+    def __init__(self, name, returncode, output):
+        self.name = name
+        self.output = output
+        self.returncode = returncode
+
+    def display_failed(self):
+        """
+        Display the output from a failed test.
+        """
+        if self.failed():
+            print("--------------------------------")
+            print("Failed test: " + self.name)
+            print(self.output)
+            print("--------------------------------")
+
+    def failed(self):
+        """
+        Returns whether or not a test failed.
+        """
+        return self.returncode != 0
+
+def load_test_file(test_file):
+    """
+    Load a given test file.
+
+    Args:
+        test_file (str): Path to the test file to load.
+    """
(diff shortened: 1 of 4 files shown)


open-ce/test_feedstock.py
```diff
@@ -0,0 +1,223 @@
...
+            print(self.output)
+            print("--------------------------------")
+
+    def failed(self):
+        """
+        Returns whether or not a test failed.
+        """
+        return self.returncode != 0
+
+def load_test_file(test_file):
```
jayfurmanek, line 127: This looks general enough to be in `utils.py` perhaps.

open-ce/test_feedstock.py
```diff
@@ -0,0 +1,223 @@
...
+        with tempfile.NamedTemporaryFile(mode='w+t', dir=os.getcwd(), delete=False) as temp:
+            temp.write("set -e\n")
+            temp.write(self.get_test_command(conda_env_file))
+            temp_file_name = temp.name
+
+        # Execute file
+        test_process = subprocess.run(["bash", temp_file_name], stdout=subprocess.PIPE, check=False,
+                                      stderr=subprocess.STDOUT, universal_newlines=True)
+
+        # Remove file containing bash commands
```
jayfurmanek, line 87: Maybe there are scenarios where we'd want this file to be written to disk? Perhaps if a test failed?

open-ce/test_feedstock.py
```diff
@@ -0,0 +1,223 @@
...
+        self.returncode = returncode
+
+    def display_failed(self):
+        """
+        Display the output from a failed test.
+        """
+        if self.failed():
+            print("--------------------------------")
+            print("Failed test: " + self.name)
+            print(self.output)
```
jayfurmanek, line 118: This output could potentially be very long. Maybe we only want the last x number of lines?

open-ce/test_feedstock.py
```diff
@@ -0,0 +1,223 @@
...
+
+def display_failed_tests(failed_tests):
+    """
+    Display a list of failed tests.
+
+    Args:
+        failed_tests (:obj:`list` of :obj:`TestResult`): A list of failed TestResult's.
+    """
+    # If any collected output, return 1, else return 0
+    if failed_tests:
```
jayfurmanek, line 191: We probably want failed test lists and full outputs to be written to disk somewhere as well.

open-ce/test_feedstock.py
```diff
@@ -0,0 +1,223 @@
...
+
+    test_commands = []
+
+    # Create conda environment for testing
+    test_commands.append(TestCommand(name="Create conda environment " + conda_env,
+                                     conda_env=conda_env,
+                                     create_env=True))
+
+    for test in test_data['tests']:
+        test_commands.append(TestCommand(name=test.get('name'), conda_env=conda_env, bash_command=test.get('command')))
```
jayfurmanek, line 159: What if the bash command needs to be a small script? Like:
 -git checkout something, 
 - mount data source, 
 - run test

It seems like we'd have to write that a one big long bash command. It'd be nice if we could write tests in `.sh` files and pass those in as well (and those cases we wouldn't have to generate a temp bash file as well)?

jayfurmanek: I see the xgboost test in https://github.com/open-ce/xgboost-feedstock/pull/10/ has multiple lines of bash commands, so that looks good.

open-ce/test_feedstock.py
```diff
@@ -0,0 +1,223 @@
...
+        print("All tests passed!")
+
+def make_parser():
+    ''' Parser for input arguments '''
+    arguments = [utils.Argument.CONDA_ENV_FILE]
+    parser = utils.make_parser(arguments, description = 'Test a feedstock as part of Open-CE')
+
+    return parser
+
+def test_feedstock(arg_strings=None):
```
jayfurmanek, line 205: Naming nit: Should we be using `test_package` here (and in the name of the file) since it really is a test on the package content and not the feedstock itself?

open-ce/test_feedstock.py
```diff
@@ -0,0 +1,223 @@
...
+    """
+    Generate a list of test commands from the provided test file.
+
+    Args:
+        test_file (str): Path to the test file.
+    """
+    test_data = load_test_file(test_file)
+
+    time_stamp = datetime.datetime.now().strftime("%Y%m%d%H%M%S")
+    conda_env = "open-ce-conda-env-" + time_stamp
```
npanpaliya, line 149: `open-ce-conda-env` is defined in utils.py. We can use that variable here.

open-ce/test_feedstock.py
bnemanich: I think this will eventually become a little more complicated where it is using variants in the same manner as the Open-CE environment files. At that point we can probably combine the two.

open-ce/test_feedstock.py
bnemanich: That might be the case, but that could end up leaving kind of a mess. I've opened an issue to choose a working directory for running tests (#139). Once that is completed, it might make sense to add an option to keep files around.

open-ce/test_feedstock.py
bnemanich: Hmm, that is an interesting question. In the XGBoost test that I set up, it is running an entire suite of pytest tests. Cutting off the output of that would probably make it hard to see what actually went wrong. I think I would prefer to see if this becomes an issue in the future once we have written more tests. If it does, we can look into ways of making it easier to parse errors.

open-ce/test_feedstock.py
bnemanich: Yep, good point.

open-ce/test_feedstock.py
bnemanich: I will open an issue for that. I think there are some existing file formats that we could output the logs in that would allow other tools to parse them better.

For now, a user can just pipe the output to a file if they are interesting in storing the output of the tests.

open-ce/test_feedstock.py
bnemanich: I'm not sure. It might even make sense to be test_repository, since these tests don't necessarily need to be associated with a particular package. I went with feedstock here since that is what we changed the name of "repository" in the Open-CE environment files to "feedstock".

I'd say what this script does is: "Run every test within a repository (or directory I suppose) inside of a given conda environment"

open-ce/test_feedstock.py
jayfurmanek: ok

open-ce/test_feedstock.py
jayfurmanek: Ok. I added a comment to #139 for this.

open-ce/test_feedstock.py
jayfurmanek: OK, makes sense. This has me thinking what if we wanted to organize tests into multiple test files ([suite1, suite2, etc] or [test1], [test2], etc] ) to help with log management and add the ability to do concurrent testing across the cluster on the same package. It would involve calling into `test_feedstock` numerous times with a different test config file each time.

open-ce/test_feedstock.py
jayfurmanek: Ok, sounds good.

open-ce/test_feedstock.py
jayfurmanek: OK. We do use `feedstock` in a bit broader context already so this is fine.

open-ce/test_feedstock.py
bnemanich: Yeah, I have a feeling we will eventually want multiple files for tests. Right now that isn't supported, but wouldn't be hard to add at some point.

jayfurmanek (approved):

PR #135, opened 2020-10-30
Validate config in build_env.

### LINKED ISSUE #80: Call `validate_config` at start of `build_env`
There could be missing runtime dependencies that won't be discovered until at the very end of a build.

We can call `validate_config` at start of `build_env`.

bethune-bryant:

## Description

Fixes #80

### DIFF
--- a/open-ce/validate_config.py
+++ b/open-ce/validate_config.py
@@ -37,7 +37,6 @@ def validate_env_config(conda_build_config, env_config_files, variants, reposito
     for a given set of variants.
     '''
     for variant in variants:
-        print('Validating {} for {}'.format(conda_build_config, variant))
         for env_file in env_config_files:
             print('Validating {} for {} : {}'.format(conda_build_config, env_file, variant))
             try:
@@ -51,10 +50,6 @@ def validate_env_config(conda_build_config, env_config_files, variants, reposito
             except OpenCEError as err:
                 raise OpenCEError(Error.VALIDATE_CONFIG, conda_build_config, env_file, variant, err.msg) from err
             print('Successfully validated {} for {} : {}'.format(conda_build_config, env_file, variant))
-
-        print('Successfully validated {} for {}'.format(conda_build_config, variant))
-
-    print("{} Successfully validated!".format(conda_build_config))
 
 def validate_build_tree(recipes, variant):
     '''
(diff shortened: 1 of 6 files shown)


open-ce/validate_config.py
```diff
@@ -26,52 +26,58 @@ def make_parser():
...
+    Validates a lits of Open-CE env files against a conda build config
+    for a given set of variants.
+    '''
     for variant in variants:
-        print('Validating {} for {}'.format(args.conda_build_config, variant))
-        for env_file in args.env_config_file:
-            print('Validating {} for {} : {}'.format(args.conda_build_config, env_file, variant))
+        print('Validating {} for {}'.format(conda_build_config, variant))
+        for env_file in env_config_files:
+            print('Validating {} for {} : {}'.format(conda_build_config, env_file, variant))
```
npanpaliya, line 42: We can remove this print statement.

npanpaliya: Or may be above one?

open-ce/validate_config.py
```diff
@@ -26,52 +26,58 @@ def make_parser():
...
-            if retval != 0:
-                raise OpenCEError(Error.VALIDATE_CONFIG, args.conda_build_config, env_file, variant, "")
+    pkg_args = " ".join(["\"{}\"".format(utils.generalize_version(dep)) for dep in deps
+                                                                    if not utils.remove_version(dep) in packages])
 
-            print('Successfully validated {} for {} : {}'.format(args.conda_build_config, env_file, variant))
+    channel_args = " ".join({"-c \"{}\"".format(channel) for channel in channels})
 
-        print('Successfully validated {} for {}'.format(args.conda_build_config, variant))
+    cli = "conda create --dry-run -n test_conda_dependencies {} {}".format(channel_args, pkg_args)
```
bnemanich, line 73: Maybe we should capture the output from this and only display it if there is an error.

open-ce/validate_config.py
bnemanich: Yeah, I think we should remove most of the print statements now that this will be run during each build.

open-ce/validate_config.py
bethune-bryant: All of this output is displayed when it is executed.
I could capture it instead and include it as part of the exception.

bethune-bryant: This is done.

open-ce/validate_config.py
bethune-bryant: I removed all but the starting and finished statements.

bethune-bryant:
@bnemanich @npanpaliya 
I cleaned up all of the printing. Let me know what you think.

bnemanich (approved):

npanpaliya:
This looks good to me now.

npanpaliya (approved):

PR #143, opened 2020-11-02
Save conda env file to output folder.

### LINKED ISSUE #126: Generated conda env files not copied out of container
**Describe the bug**
When using the `--docker_build` command, the conda environment files are not copied out of the container.

**To Reproduce**
Build any Open-CE environment using the `--docker_build` command. Towards the end of the output, you will see something like: `Generated conda environment files from the selected build arguments: ['/home/builder/opence-conda-env-py3.6-cpu-openmpi.yaml', '/home/builder/opence-conda-env-py3.6-cuda-openmpi.yaml']`

These files are only generated within the container.

**Expected behavior**
The generated conda environment files should be copied out of the container. They should be stored within the current working directory.

bethune-bryant:

## Description

Fixes #126 

I changed the output location for the conda env files to be the build output directory. This ensures that it is available outside of the docker container after a build.

### DIFF
--- a/open-ce/build_env.py
+++ b/open-ce/build_env.py
@@ -34,6 +34,7 @@ For usage description of arguments, this script supports use of --help:
 
 import os
 import sys
+import glob
 
 import build_feedstock
 import docker_build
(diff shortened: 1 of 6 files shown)


jayfurmanek (approved):
Good idea!

open-ce/build_env.py
```diff
@@ -123,7 +123,7 @@ def build_env(arg_strings=None):
 
         conda_env_data.update_conda_env_file_content(build_command, build_tree)
 
-    conda_env_files = conda_env_data.write_conda_env_files()
+    conda_env_files = conda_env_data.write_conda_env_files(args.output_folder)
```
npanpaliya, line 126: I think we should probably change the default argument of `write_conda_env_files` to DEFAULT_OUTPUT_FOLDER instead of os.getcwd().

npanpaliya (approved):

bnemanich:
Can you verify that the local channel that appears in the conda environment file has the correct path?

npanpaliya:
> Can you verify that the local channel that appears in the conda environment file has the correct path?

That local channel path points to output_folder and it is absolute. So, it should be intact even after this change.

bethune-bryant:
> 
> 
> Can you verify that the local channel that appears in the conda environment file has the correct path?

Good catch. It has the output path for within the container:
```
channels:
- file://home/builder/condabuild
```

I'll look into what to do about this.

bethune-bryant:
This should be good to go now.
I also changed the docker_build option to always use ~/condabuild within the container. I couldn't see a reason to need to specify a different internal output directory, and it was causing a lot of issues.

npanpaliya (approved):

bnemanich (approved):

PR #162, opened 2020-11-06
Update dependencies for Tensorflow2.4 and PyTorch 1.7

jayfurmanek:
Working PR to update dependency pins for Tensorflow 2.4

Of the below, all are available in defaults except `opt_einsum 3.3.0` so we'll have to watch that one.


jayfurmanek:
Updated for grpcio. Latest on defaults is 1.31 at the moment.

npanpaliya:
Even if I reset opt_einsum and grpcio pins to back earlier versions, I can't get these pins validated for tensorflow-env.yaml on both x86 and ppc.

npanpaliya:
I got the cause of the conflicts. tensorflow-estimator needs to use the right pin for `typing_extensions`. Please see this https://github.com/open-ce/tensorflow-estimator-feedstock/pull/5/files#r521319452.

conda_build_config.yaml
```diff
@@ -42,29 +42,39 @@ cython:
...
 google_pasta:
-  - 0.1.8
+  - 0.2
 googleapis_common_protos:
   - 1.52.*
 graphsurgeon:
   - 0.4.1
 grpcio:
-  - 1.16.1
+  - 1.32.*
```
npanpaliya, line 77: @jayfurmanek - I think we should change the versions of grpcio and opt_einsum to whatever latest is available which are 1.31.* and 3.1.0 respectively. Otherwise, we won't be able to get TF and dependent environment builds. We can always update these to newer versions when they are available.

npanpaliya:
@jayfurmanek - I think we need to update/remove the pins for cudnn and cudatoolkit from here. Because when we try to build any environment for cuda_versions=11.0, then the needed tensorrt is 7.2 which needs cudnn8 and cudatoolkit should be 11.0.  Or I think, we should have something like this -
```
cudnn:
  - 7.6     #[cuda_version == 10.2]
  - 8.*      #[cuda_version == 11.*]
```
But I don't know if it works. I'll try it.

jayfurmanek:
The PR to update cuDNN to version 8 (https://github.com/open-ce/cudnn-feedstock/pull/6) specifies version `8.0.4.30-1` for both CUDA 10.2 and CUDA 11.0. 

NVIDIA does make cuDNN8 available for both cuda10.2 and 11 on both x86-64 and ppc64le. However I do have some concerns with Tensorflow. These combinations are not officially listed on the TF compatibility matrix: https://www.tensorflow.org/install/source#gpu

But CUDA10.2 is not on there at all in any capacity either and we know that works ok.

jayfurmanek:
Protobuf update dependent on what TensorFlow does with: https://github.com/tensorflow/tensorflow/issues/45300

Issue #88, opened 2020-10-12
Verify that system CUDA version matches cudatoolkit version


bnemanich:
The build_feedstock script should check that the version of CUDA installed matches the version that is provided by the cudatoolkit version. If it doesn't match, an error should be thrown and the build should fail.

It should also check that CUDA_HOME is set. If it isn't set, it can be set with a default value.

It should also check that CUDA_HOME matches the `nvcc` that is in the system PATH.

dncliss:
Nothing much on my plate at the moment while waiting on some older PRs to be reviewed -- I'll self-assign and take look at this one.

dncliss:
This one got set aside 3 months ago.  Re-engaged; Killed earlier outdated PR and started fresh.

dncliss:
The earlier attempt to do this item remains available, in the PR mentioned above.  But I never finished it off and I think it's best that I put it back into the backlog pool (either for someone else to pick up, or to dispose if no longer desired.)

alexlang74:
No intent to address

PR #171, opened 2020-11-11
Requirement pt 1: check cudatoolkit (nvcc) against system CUDA version

dncliss:

- [ x] Did you read the [contributor guide](https://github.com/open-ce/open-ce/blob/master/CONTRIBUTING.md)?

## Description
Two things:
1) Ensure that system CUDA version matches cudatoolkit (nvcc) version.
2) Check CUDA_HOME is set; if not, default it to match parent dirname of nvcc.
Will push as two commits in this same PR, as both items are for same issue.

Fixes # (issue).  https://github.com/open-ce/open-ce/issues/88

## Review process to land 

### DIFF
--- a/open-ce/build_feedstock.py
+++ b/open-ce/build_feedstock.py
@@ -147,6 +147,15 @@ def _set_local_src_dir(local_src_dir_arg, recipe, recipe_config_file):
         if 'LOCAL_SRC_DIR' in os.environ:
             del os.environ['LOCAL_SRC_DIR']
 
+def check_cuda_version_match():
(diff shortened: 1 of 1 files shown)


dncliss:
Oh ... I should note:  don't review this yet.

I'm pushing the two requirements under two commits, and this was the first one.
Second part to come shortly....

jayfurmanek:
opps! already took a look. :)
Remember we have some cuda level checking function in utils.py that we use during docker builds. Let's reuse, or add to, those if possible.

Issue #187, opened 2020-11-19
[BUG] git_tag specified for the feedstock is not being used


npanpaliya:
**Describe the bug**
git_tag specified for the feedstock in the Open-CE environment file is not being used. Ideally it should be respected over git_tag_for_env.

**To Reproduce**
1. Specify git_tag for any feedstock in any of the Open-CE environment file
2. Also specify git_tag_for_env in the environment file

**Expected behavior**
The feedstock for which `git_tag` is specified should clone that particular git tag/branch/version of the feedstock instead of the one specified as `git_tag_for_env`.

PR #190, opened 2020-11-23
Add patches support

### LINKED ISSUE #186: [REQUEST] Add patches field in Open-CE environment files
Yes. In order to build external recipes like from conda-forge or Anaconda, our current build scripts work fine. However, in case the recipe needs some change to work for ppc64le or in our setup, this feature will facilitate that.

This request is about adding `patches` field in the Open-CE environment files for each feedstock. This field will be optional. Patches to be applied should be placed in github.com/open-ce/open-ce repository.

npanpaliya:

## Description

Fixes #186  and #187 .

Documentation update is being done. Once that is done, will update this PR.

### DIFF
--- a/open-ce/build_tree.py
+++ b/open-ce/build_tree.py
@@ -322,6 +322,7 @@ class BuildTree(): #pylint: disable=too-many-instance-attributes
 
         return build_commands, external_deps, test_commands
 
+    #pylint: disable=too-many-branches
     def _clone_repo(self, git_url, repo_dir, env_config_data, package):
         """
         Clone the git repo at repository.
--- a/open-ce/build_tree.py
+++ b/open-ce/build_tree.py
@@ -340,34 +340,22 @@ class BuildTree(): #pylint: disable=too-many-instance-attributes
             else:
                 git_tag = env_config_data.get(env_config.Key.git_tag_for_env.name, None) if env_config_data else None
 
-        if git_tag is None:
-            clone_cmd = "git clone " + git_url + " " + repo_dir
-        else:
-            clone_cmd = "git clone -b " + git_tag + " --single-branch " + git_url + " " + repo_dir
-
+        clone_cmd = "git clone " + git_url + " " + repo_dir
         print("Clone cmd: ", clone_cmd)
         clone_result = os.system(clone_cmd)
+
         cur_dir = os.getcwd()
         clone_successful = clone_result == 0
-        if not clone_successful:
+        if clone_successful:
             if not git_tag is None:
-                # If above clone command failed and git tag was specified,
-                # then retry git clone and git checkout for the given git_tag.
-                # This would be the case when git_tag specified is a commit hash
-                # instead of git tag or branch.
-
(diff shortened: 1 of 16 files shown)


npanpaliya:
Depends on https://github.com/open-ce/LightGBM-feedstock/pull/1

open-ce/build_tree.py
```diff
@@ -345,8 +347,44 @@ def _clone_repo(self, git_url, repo_dir, env_config_data, git_tag_from_config):
 
         print("Clone cmd: ", clone_cmd)
         clone_result = os.system(clone_cmd)
-        if clone_result != 0:
-            raise OpenCEError(Error.CLONE_REPO, git_url)
+        cur_dir = os.getcwd()
+        clone_successful = clone_result == 0
+        if not clone_successful:
+            if not git_tag is None:
```
jayfurmanek, line 353: Is there a better way than catching a failed clone?
Maybe we could determine if its a commit hash ahead of time above, then make sure we issue the correct `clone` or `clone` + `checkout` commands?

jayfurmanek:
This looks pretty good. I had one comment.

open-ce/build_tree.py
npanpaliya: I searched for git documentation/references where this can be checked. But unfortunately, without cloning the repo, we can't really get to know if the specified value under `git tag` is a branch/tag name or a commit hash. Also, branch name can have as many characters as that of a commit hash (For e.g. we can create branch name with `499b5b28e0a349b9c351881224e34cfaacc463fd` which is actually a commit hash). 
One way is we add another field `commit_hash` in our environment file for it as an alternative to `git_tag`.  So, if `commit_hash` is specified, we can do clone + checkout.

bnemanich: Couldn't we always do this step, and just skip using the `-b` option all together?

npanpaliya: Yes, we can do.

tests/helpers.py
```diff
@@ -24,19 +24,24 @@ def validate_cli(cli_string, expect=None, reject=None, ignore=None, retval=0):
         expect: A list of strings that must occur in the 'cli_string' arg.
         reject: A list of strings that cannot occur in the 'cli_string' arg.
         ignore: Don't validate the CLI_STRING if one of these strings is contained in it.
+        possible_expect: A list of possible strings that the CLI may contain. CLI should contain
+                        atleast one of these.
```
bnemanich, line 28: atleast should be "at least"

bnemanich:
Looks good. I added on small comment for a typo in a comment.

This won't pass tests until the LightGBM feedstock is finished. You could take the environment file out for now.

If you are waiting until LightGBM is finished, you should also add the LightGBM environment file to the main Open-CE environment file.

open-ce/build_tree.py
jayfurmanek: ok, this is better.

jayfurmanek (approved):
Looks good. Thanks @npanpaliya !

bnemanich (approved):

PR #213, opened 2020-12-09
Correct validation in build_env.

### LINKED ISSUE #211: [BUG] During `build env` Validation is always performed on default branch
**Describe the bug**
When performing the `open-ce build env` action, validation happens before the BuildTree is constructed, so it clones its own copies of the repo.

**To Reproduce**
Run `build env` with a non default git_tag.

**Expected behavior**
Validation is performed on the specified tag.

bethune-bryant:

## Description

Fixes #211

### DIFF
--- a/open-ce/validate_config.py
+++ b/open-ce/validate_config.py
@@ -39,13 +39,13 @@ def validate_env_config(conda_build_config, env_config_files, variants, reposito
         for env_file in env_config_files:
             print('Validating {} for {} : {}'.format(conda_build_config, env_file, variant))
             try:
-                recipes = build_tree.BuildTree([env_file],
-                                               variant['python'],
-                                               variant['build_type'],
-                                               variant['mpi_type'],
-                                               variant['cudatoolkit'],
-                                               repository_folder=repository_folder,
-                                               conda_build_config=conda_build_config)
+                _ = build_tree.BuildTree([env_file],
+                                          variant['python'],
+                                          variant['build_type'],
+                                          variant['mpi_type'],
+                                          variant['cudatoolkit'],
+                                          repository_folder=repository_folder,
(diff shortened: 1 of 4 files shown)


bethune-bryant:
@bnemanich 
I had to change it a little after we talked about it.

npanpaliya (approved):

open-ce/validate_config.py
```diff
@@ -46,19 +46,19 @@ def validate_env_config(conda_build_config, env_config_files, variants, reposito
                                                variant['cudatoolkit'],
                                                repository_folder=repository_folder,
                                                conda_build_config=conda_build_config)
-                validate_build_tree(recipes, variant)
+                validate_build_tree(recipes, recipes.get_external_dependencies(variant))
```
bnemanich, line 49: Do we still need to call validated_build_tree here? It seems like we are calling it twice now. Once in the BuildTree constructor, and once here.

bethune-bryant: You're right. I didn't commit the change that removed this.
Thanks for catching that!

bethune-bryant: Actually, this is the call for the validate_config entry point.
The build_tree constructor calls the validate build tree function.

bnemanich: It looks like BuildTree's constructor will call `validate_config.validate_build_tree`. The constructor is called right before this line. This line then calls `validate_config.validate_build_tree` again.

bnemanich: The tests seem to be taking twice as long as before as well.

bethune-bryant: So the entry_point path was doing a double call now that the validation is happening as part of the build_tree constructor. That should be fixed now.

bnemanich (approved):

PR #215, opened 2020-12-10
Update check_build_numbers.py for cuda_versions

bethune-bryant:
Blocking open-ce/cudnn-feedstock#6


bethune-bryant:
@jayfurmanek 
This should fix the issue you were seeing in open-ce/cudnn-feedstock#6.
Once this is merged we can update that issue.

bnemanich (approved):

Issue #194, opened 2020-12-01
Cleanup Generated Conda Env File


bethune-bryant:
A user has stated that the generated Conda Env file is confusing because there are duplicate packages in arbitrary places.

I think a simple first step to making the file easier to view is to sort the packages.

If there is any other simple cleanup that we can do it may be good to do that as well. Maybe removing explicit duplicates if there are any.

bnemanich:
This was completed by #220 .

PR #220, opened 2020-12-14
Removed build time packages and duplicates from conda environment file

### LINKED ISSUE #212: [REQUEST] Add a way to keep packages out of Conda Env file
Packages that are only needed at build time are being included in the generated conda env files.

We can add a property to the feedstock object in the open-ce env files called `runtime_package` that defaults to True. This value being false will keep it out of the generated conda env file.

Non runtime packages should also be removed when performing validation to ensure that there isn't something needed that won't be included.

**Additional context**
Example:
```
packages:
  - feedstock : cdts
    runtime_package : False
  - feedstock : bazel
    runtime_package : False
```

Related to #194

npanpaliya:

## Description

Fixes #212 and #194 .


npanpaliya:
I'm working on tests, while the main logic is being reviewed. Also, it would be good if I get some early feedback on this.

open-ce/validate_config.py
```diff
@@ -46,27 +46,28 @@ def validate_env_config(conda_build_config, env_config_files, variants, reposito
...
     '''
     Check a build tree for dependency compatability.
     '''
-    packages = [package for recipe in recipes for package in recipe.packages]
     channels = {channel for recipe in recipes for channel in recipe.channels}
-    deps = {dep for recipe in recipes for dep in recipe.run_dependencies}
-    deps.update(recipes.get_external_dependencies(variant))
 
-    pkg_args = " ".join(["\"{}\"".format(utils.generalize_version(dep)) for dep in deps
-                                                                    if not utils.remove_version(dep) in packages])
```
npanpaliya, line 64: While working on tests, I found that one of the tests from validate_config_test.py is failing because we don't expect the package from env file to be present in the cli_string while verifying `conda create --dry-run... `. I think the main package itself should always be there while environment creation. Is validate_config meant to verify only dependencies's installation in an env or the main package along with its dependencies?

npanpaliya: Reverting above change fixes the test.

open-ce/build_tree.py
```diff
@@ -443,6 +449,59 @@ def get_external_dependencies(self, variant):
...
+                dep_name_from_set = dep.split()[0]
+                if dep_name_to_be_added == dep_name_from_set:
+                    return None
+
+            # If no match found, just add it
+            return dep_to_be_added
+
+        def _get_unique_deps_names(dependencies):
+            deps = set()
+            if not dependencies is None:
```
bnemanich, line 479: This can just be `if dependencies`

open-ce/build_tree.py
```diff
@@ -443,6 +449,59 @@ def get_external_dependencies(self, variant):
...
+            dependencies = _get_unique_deps_names(dependencies)
+
+            for dep in dependencies:
+                pack_to_add = check_matching(parent_set, dep)
+                if pack_to_add:
+                    parent_set.add(pack_to_add)
+
+            return parent_set
+
+        for build_command in self.build_commands:
```
bnemanich, line 497: I'm not sure if this will work. This will probably contain the dependencies for every variant that has been seen up to this point.

bnemanich: You'll probably need to keep the CondaEnvFileGenerator call where it was previously, and still pass in the build_commands.

npanpaliya: I'll try that. But so far, I've tested TF/dali/lightGBM recipes and environment files created for them. They were all working. Do you think moving the CondaEnvFileGenerator call where it was will reduce the number of dependencies (duplicate ones)?

npanpaliya: I think now I understand what you're saying. I'll make the changes.

bnemanich:
Also, can you please update the documentation to account for the new `runtime_package` flag in Open-CE Environment files.

bnemanich (approved):

npanpaliya:
@jayfurmanek - Brad has approved this. If you have no more comments on this, can I merge this one?

jayfurmanek:
Yes, sorry. Let's go head and merge. Thanks!

jayfurmanek (approved):

PR #238, opened 2021-01-04
Add the create_version_branch script

bnemanich:

## Description

Adds a script that will:
* Create a new branch within a feedstock whose name is based on the version number of the feedstock's package.

This script should be run on the commit just before the version of the feedstock's packages are incremented. At the moment it will have to be run manually. At some point, we can look into using GitHub's actions to run it on every push to the master branch of a feedstock.

## Review process to land 

### DIFF
--- a/git_tools/create_version_branch.py
+++ b/git_tools/create_version_branch.py
@@ -72,7 +72,7 @@ def _get_repo_name(repo_url):
         repo_url += ".git"
     return os.path.splitext(os.path.basename(repo_url))[0]
 
-def _main(arg_strings=None):
+def _create_version_branch(arg_strings=None):
     parser = _make_parser()
     args = parser.parse_args(arg_strings)
 
@@ -100,7 +100,7 @@ def _main(arg_strings=None):
 
 if __name__ == '__main__':
     try:
-        _main()
+        _create_version_branch()
     except Exception as exc:# pylint: disable=broad-except
         print("Error: ", exc)
         sys.exit(1)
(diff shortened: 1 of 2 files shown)


jayfurmanek (approved):
This should be helpful! Looks good.

git_tools/create_version_branch.py
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env python
+"""
+*****************************************************************
+Licensed Materials - Property of IBM
+(C) Copyright IBM Corp. 2020. All Rights Reserved.
```
npanpaliya, line 5: This needs to be 2021.

git_tools/create_version_branch.py
```diff
@@ -0,0 +1,106 @@
...
+    build_config_data, _ = build_feedstock.load_package_config()
+
+    for recipe in build_config_data["recipes"]:
+        rendered_recipe = conda_utils.render_yaml(recipe["path"])
+        for meta, _, _ in rendered_recipe:
+            package_version = meta.meta['package']['version']
+            if package_version and package_version != "None":
+                os.chdir(saved_working_directory)
+                return package_version
+
```
npanpaliya, line 67: Should changing back to saved_working_directory be done here too or may be break the look instead of returning from here? If for some reason, package_version isn't found or is "None", the current directory won't be restored back.

npanpaliya (approved):

cdeepali (approved):

bethune-bryant (approved):

Issue #236, opened 2020-12-28
CuDNN 8 on X86 missing memcpy reference


bnemanich:
@cdeepali reported seeing the following error while building PyTorch for PYthon 3.8 on x86:

```
/home/deepch23/anaconda3/conda-bld/pytorch-base_1608825300544/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/lib64/libcudnn.so: undefined reference to `memcpy@GLIBC_2.14'
collect2: error: ld returned 1 exit status
```

It seems like CuDNN 8 on x86 also needs the memcpy_wrapper that is currently part of the TensorRT feedstock. This is only a problem on Python 3.8, because TensorRT is not being used with PyTorch with Python 3.8.

The memcpy_wrapper should be moved from the TensorRT feedstock to the CuDNN feedstock. It will still get pulled into TensorRT since CuDNN is a dependency for TensorRT.

npanpaliya:
I think we should put those memcpy and clocktime wrappers to a new and separate repository.

jayfurmanek:
@cdeepali In your build tests, did you move the wrapper to it's own repo, or was it in the cudnn repo?

cdeepali:
@jayfurmanek, I just had local changes to enable it. I had not moved it it was under TRT itself.

PR #239, opened 2021-01-05
add libc_wrappers

cdeepali:

## Description

Fixes https://github.com/open-ce/open-ce/issues/236.

## Review process to land 

### DIFF
--- a/envs/tensorrt-env.yaml
+++ b/envs/tensorrt-env.yaml
@@ -1,8 +1,11 @@
 packages:
 {% set py_ver = python | float %}
-{% if build_type == 'cuda' and py_ver < 3.8 %}
+{% if build_type == 'cuda' %}
+  - feedstock: libc_wrappers     #[cudatoolkit == '11.0']
+{% if py_ver < 3.8 %}
   - feedstock : cudnn
   - feedstock : tensorrt
   - feedstock : onnx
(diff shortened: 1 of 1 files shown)


envs/tensorrt-env.yaml
```diff
@@ -1,8 +1,11 @@
 packages:
 {% set py_ver = python | float %}
-{% if build_type == 'cuda' and py_ver < 3.8 %}
+{% if build_type == 'cuda' %}
```
npanpaliya, line 3: Approved this if cpu builds work fine with this.

cdeepali: Yes, cpu builds are fine.

bnemanich: We might still need to add libc_wrappers to other environment files. Are there any environment files that include CuDNN, but not TensorRT?

cdeepali: no change in cudnn repo for libc_wrappers.

jayfurmanek: Currently there is no other env file that uses cudnn (and not TRT), but I suppose there could be. Or someone could create a custom one, in which case they'd have to know to put in the wrappers.

jayfurmanek (approved):

PR #276, opened 2021-01-22
Add the recipe_path field to open-ce environment files

bnemanich:

## Description

This PR does two things that make it easier to build external feedstocks.

1. Adds the `recipe_path` field to a `package` within an Open-CE Environment file. This allows for external feedstocks that don't keep their recipes within a directory called `recipe`.
2. Makes all dependencies lower-case internally. I found some recipes that had packages with upper-case letters. `conda` treats these the same as if they were all lower-case, so the Open-CE tools should do the same.

## Review process to land 

### DIFF
--- a/doc/README.yaml.md
+++ b/doc/README.yaml.md
@@ -33,16 +33,17 @@ serving a different purpose. All are used to define key-value pairs as used in a
 YAML file format. The keywords recognized for Open-CE environments include the
 following:
 ```
-packages:      # The environment package name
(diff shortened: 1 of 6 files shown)


npanpaliya:
We also want to update documentation with this new field. I think there should also be a test that covers the case of this field being read from environment file.

bnemanich:
Good point @npanpaliya . I added a new commit with some documentation and another test.

npanpaliya (approved):
Thanks @bnemanich . Looks good to me now.

PR #278, opened 2021-01-22
Separate Env Files

### LINKED ISSUE #257: Move Open-CE build tools to a separate repo from the Open-CE environment files.
The Open-CE build tools and default Open-CE environment files should be kept within separate repositories, and versioned differently.

This issue should:
1. Determine the naming of the repositories.
2. Move the appropriate code to the approrpiate repositories.
3. Create plan for keeping Open-CE environment files backwards compatible with future Open-CE build tools.
4. Determine versioning scheme for Open-CE build tools.

bethune-bryant:

## Description

Fixes #257 

~~blocked by open-ce/open-ce-environments#1~~


bnemanich:
We will have to think about a default location for the `conda_build_config.yaml` file. I don't think we want someone to always have to pass an argument in to specify where it will be. Currently the default location is relative to the build tools. Instead, it should probably be relative to the environment files. Another option is to have the environment files specify where to look for the `conda_build_config.yaml` files.

doc/README.yaml.md
```diff
@@ -1,10 +1,9 @@
 
 # README for Open-CE Environment Files
 
-
 This README documents the Open-CE Environment YAML files used by the build scripts
-[build_env.py](README.build_env.md)
-and [build_feedstock.py](README.build_feedstock.md).
+[open-ce build env](hREADME.open_ce_build.md#open-ce-build-env-sub-command)
```
jayfurmanek, line 5: extra 'h' here

bethune-bryant: Good catch. Thanks!

README.md
```diff
@@ -70,24 +70,28 @@ When building packages that use CUDA, a tar package of TensorRT for the intended
...
+The following commands will use the `opence-env.yaml` Open-CE environment file from a specific Open-CE release to build all of the Open-CE packages for Python 3.6, 3.7 and 3.8, including only CUDA builds. The commands should be run from within the same directory that contains `local_files`.
 
 ```bash
 # Clone a specific Open-CE release from GitHub
 git clone https://github.com/open-ce/open-ce.git --branch open-ce-v1.0.0
+# Clone Open-CE Environments from GitHub
+git clone https://github.com/open-ce/open-ce-environments.git --branch open-ce-v2021.1.0
 # Build packages
-./open-ce/open-ce/open-ce build env --python_versions 3.6,3.7,3.8 --build_types cuda open-ce/envs/opence-env.yaml
+./open-ce/open-ce/open-ce build env --python_versions 3.6,3.7,3.8 --build_types cuda ./open-ce-environments/envs/opence-env.yaml
```
bnemanich, line 94: Will these commands work with the `--docker_build` option. Will the environment files get passed into the container?

bethune-bryant: No it won't. I forgot to look into that.
We may need to add another volume mount to the docker command.

bethune-bryant: I got around it during some testing by cloning the environments repo into the open-ce folder.

bethune-bryant: But with the absolute path conversion being handled earlier on, it won't work. I'll need to look into this more.

bnemanich: Copying the environment files into the container is probably the best thing to do. That actually makes things a little more flexible than what we currently have as well. Probably need to copy in the environment files and the conda_build_config file.

bethune-bryant: The only problem I see with this is that the env file imports other env files. Should we:

1.  fully resolve all imports and copy all those files in
2.  just copy all yaml files in the same folder as the env file
3.  copy all contents of the folder that contians the env file

Also, I don't recall. Can env files import files from anywhere in the file system?
I imagine if so, then dockerbuild doesn't work in general as it is now, so should we even worry about that?

jayfurmanek: yeah, it looks like if it's given a full path anywhere, it'll try and load it, otherwise it looks in the standard location:

https://github.ibm.com/open-ce/open-ce/blob/master/open-ce/env_config.py#L91-L96

With this new model of the environments being separate, maybe we want to change it so it only looks in the specified open-ce environments repo. That would be more deterministic (and more secure). And I guess that would mean your option 3 (or I guess option 2 as well if we wanted to be more specific with it).

bethune-bryant: I am now mounting the directory that contains the env files, and copying the conda build config into the container.

Regarding the env files importing from anywhere in the fs, this method has the same limitations the previous implementation had, so I think if we want to change this behavior, we change it in another PR.

open-ce/docker_build.py
```diff
@@ -121,37 +126,48 @@ def _stop_container(container_name):
...
     time_stamp = datetime.datetime.now().strftime("%Y%m%d%H%M%S")
     container_name = IMAGE_NAME + "-" + time_stamp
 
-    _create_container(container_name, image_name, output_folder)
+    output_folder = os.path.abspath(args.output_folder)
+    env_file = os.path.abspath(args.env_config_file[0])
+    conda_build_config = os.path.abspath(args.conda_build_config)
+
+    env_folder = os.path.dirname(env_file)
+    env_file_in_container = os.path.join(HOME_PATH, "envs", os.path.basename(env_file))
```
bnemanich, line 141: I think we need to do this for every file in the args.env_config_file list.

bethune-bryant: There can't be more than one.

https://github.com/open-ce/open-ce/blob/1443d456b755c131c90892ac13c927be098aaf14/open-ce/inputs.py#L57-L63

bethune-bryant: It used to allow more than one. Was that something that was intended?

bnemanich: You can still pass multiple environment files. I think that `nargs=1` just means the number of arguments to be consumed.

The help indicates this as well, since the usage is outputted as: `env_config_file [env_config_file ...]`

bethune-bryant: This is what I get with my current changes:
```
(dev) [bnelson@dlw05 open-ce]$ ./open-ce/open-ce build env ../open-ce-environments/envs/xgboost-env.yaml ../open-ce-environments/envs/spacy-env.yaml --docker_build --python_versions 3.7 --build_types cuda
usage: open-ce [-h] {build,test,validate} ...
open-ce: error: unrecognized arguments: ../open-ce-environments/envs/spacy-env.yaml
```

bethune-bryant:
```
(dev) [bnelson@dlw05 open-ce]$ ./open-ce/open-ce build env -h
usage: open-ce build env [-h] [--conda_build_config CONDA_BUILD_CONFIG]
                         [--output_folder OUTPUT_FOLDER]
                         [--channels CHANNELS_LIST]
                         [--repository_folder REPOSITORY_FOLDER]
                         [--python_versions PYTHON_VERSIONS]
                         [--build_types BUILD_TYPES] [--mpi_types MPI_TYPES]
                         [--cuda_versions CUDA_VERSIONS]
                         [--skip_build_packages] [--run_tests]
                         [--docker_build] [--git_location GIT_LOCATION]
                         [--git_tag_for_env GIT_TAG_FOR_ENV]
                         [--test_labels TEST_LABELS]
                         env_config_file

positional arguments:
  env_config_file       Environment config file. This should be a YAML file
                        describing the package environment you wish to build.
                        A collection of files exist under the envs directory.

optional arguments:
...
```

bethune-bryant: I may have messed that up.
It looks like the other parts of the tools do support that being multiple files. I just never used that.

Let me look at changing it back.

bethune-bryant: @bnemanich 
I added this back in, but it has a limitation that there will be a conflict if env files reside in different directories with the same name.
I'm looking at a way to get around that. Maybe instead of using the directory name as the mount point, I could do a concatenation of the full path without the slashes.

bethune-bryant: I've handled this limitation. It should be good to go now.

open-ce/docker_build.py
```diff
@@ -80,7 +82,12 @@ def _add_volume(local_path, container_path):
 
     return volume_arg
 
-def _create_container(container_name, image_name, output_folder):
+def _mount_name(folder_path):
+    if len(folder_path) <= 1:
+        return ""
+    return _mount_name(os.path.dirname(folder_path)) + os.path.basename(folder_path)
+
+def _create_container(container_name, image_name, output_folder, env_folders):
```
bethune-bryant, line 90: @bnemanich 
I'm using this function to generate a name for each env file mount point.

bnemanich (approved):

jayfurmanek (approved):

Issue #127, opened 2020-10-27
[REQUEST] Provide mechanism to pass in docker build environmental variables


dllehr81:
There are cases when a user may want a docker build to have additional custom environmental variables.  For example, by default, nvidia's docker hooks, pass in all gpus to the container at runtime.  A client may wish to limit this to one, two, or maybe none(which will disable the hook altogether).  Currently I don't believe there is a way to do this?

Ideally this would be something that can be passed in on the `build_env.py` command line and passed along to the `docker build` command.

We could have it set as ENV variables if needed as well.

Issue #291, opened 2021-02-01
[BUG] Unclear description of --docker_build option


rolweber:
The help text for `--docker_build` currently says...
```txt
NOTE: When the --docker_build flag is used, all arguments with paths
      should be relative to the directory containing open-ce.
```
The command for starting a build as documented [in README.md](https://github.com/open-ce/open-ce/blob/master/README.md#building-a-collection-of-packages) is:
```bash
./open-ce/open-ce/open-ce build env open-ce/envs/opence-env.yaml
```
To which of these three directory levels does "open-ce" in the help text refer?

[In README.open_ce_build.txt](https://github.com/open-ce/open-ce/blob/master/doc/README.open_ce_build.md#open-ce-build-env-sub-command), there are still two levels to choose from:
```bash
./open-ce/open-ce build env envs/tensorflow-env.yaml
```

npanpaliya:
`open-ce`  referred in the doc refers to the root open-ce folder which comes when we clone https://github.com/open-ce/open-ce and not the sub-directories of cloned open-ce repo. 
We will precisely mention that in the doc.

npanpaliya:
This has been fixed in https://github.com/open-ce/open-ce/pull/296

PR #296, opened 2021-02-04
Docker build env vars

npanpaliya:

## Description

Fixes https://github.com/open-ce/open-ce/issues/127.

I'll update the documentation while this is being reviewed.

## Review process to land 


npanpaliya:
I've updated the documentation as reported in https://github.com/open-ce/open-ce/issues/291.

open-ce/docker_build.py
```diff
@@ -140,8 +154,8 @@ def build_in_container(image_name, output_folder, arg_strings):
     if os.path.isdir(LOCAL_FILES_PATH):
         _copy_to_container(LOCAL_FILES_PATH, HOME_PATH, container_name)
 
-    _start_container(container_name)
```
jayfurmanek, line 157: looks like an extra line was added here that's not needed.

jayfurmanek:
This looks good! It will need to account for the new conflicts.
A good way to test this function would be trying to pass in CUDA_VISIBLE_DEVICES.

bethune-bryant:
Are there any other docker options that people may want to use?

Should we generalize this to just be `--docker_build_args`, and then the user could do:
```
./open-ce/open-ce build env --docker_build --docker_build_args="--build-arg ENV1=test1 --build-arg ENV2=test2" envs/pytorch-env.yaml
```
And they could also do more generic things like:
```
./open-ce/open-ce build env --docker_build --docker_build_args="--build-arg ENV1=test1 --cpuset-cpus 0,1" envs/pytorch-env.yaml
```

@bnemanich @jayfurmanek

bnemanich:
Yes, I think what @bethune-bryant mentioned makes sense.

npanpaliya:
@bethune-bryant - This makes sense to me too. I'll incorporate this.

npanpaliya:
I'll update the documentation once the implementation is approved.

bethune-bryant:
This looks good to me now!

open-ce/docker_build.py
npanpaliya: Corrected.

bnemanich (approved):

cdeepali (approved):

PR #312, opened 2021-02-10
Add the packages argument

### LINKED ISSUE #304: [REQUEST] Generate Filtered Dependency List
For users who want to have a list of dependencies for a subset of packages it would be useful to have a feature that generates a list of dependencies for a given list of packages.

bnemanich:

## Description

Fixes #304 .

Adds the `--packages` argument to `build env`. This allows a user to only build a subset of packages from within an Open-CE environment file.

This issue also cleans up the generated conda environment files. It only adds packages that are being built (not their direct dependencies). It also adds the version number to all of the packages that are being built.

## Review process to land 

### DIFF
--- a/open-ce/build_tree.py
+++ b/open-ce/build_tree.py
@@ -35,6 +35,7 @@ class BuildCommand():
                  recipe,
                  repository,
                  packages,
+                 version=None,
                  recipe_path=None,
                  runtime_package=True,
                  output_files=None,
@@ -51,6 +52,7 @@ class BuildCommand():
         self.recipe = recipe
         self.repository = repository
         self.packages = packages
+        self.version = version
         self.recipe_path = recipe_path
         self.runtime_package = runtime_package
         self.output_files = output_files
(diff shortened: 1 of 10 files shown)


tests/build_env_test.py
```diff
@@ -215,6 +225,36 @@ def test_build_env(mocker, capsys):
     open_ce._main(["build", build_env.COMMAND, env_file, "--cuda_versions", cuda_version])
     validate_conda_env_files(cuda_versions=cuda_version)
 
+    #---The seventh test specifices specific packages that should be built (plus their dependencies)
```
npanpaliya, line 228: Minor typo - specifies

open-ce/build_tree.py
```diff
@@ -293,32 +316,53 @@ def __init__(self,
...
+                    else:
+                        print("INFO: No recipes were found for " + package + " for variant " + variant_string)
+            else:
+                for package in package_indices:
+                    variant_package_indices += package_indices[package]
+            self._initial_package_indices += variant_package_indices
 
-        self._detect_cycle()
+            validate_config.validate_build_tree(self.build_commands, external_deps, variant_package_indices)
+            installable_packages = get_installable_packages(self.build_commands, external_deps, variant_package_indices)
```
npanpaliya, line 356: Should it be just `build_commands` here? `self.build_commands` will have all of the commands for all the variants seen till now. I'm not very sure, but may be a point to re-think.

bnemanich: I am purposely using `self.build_commands` here because `variant_package_indices` might have indices that point to build_commands that were generated by previous variants. This is because there might be duplicate packages between variants. When that is the case, we use the original one that was added, and don't add the newer duplicates.

npanpaliya: Okay.

jayfurmanek (approved):

PR #319, opened 2021-02-18
Add CUDA version match check; Add CUDA_HOME path check

### LINKED ISSUE #88: Verify that system CUDA version matches cudatoolkit version

dncliss:

## Description

Fixes #88

## Review process to land 

Can't currently test due to temporary system access issue -- thus marking as draft PR for now. 
(Initial test before PR was performed with fake simulated commands to validate correctness).


dncliss:
Just wanted to note an implementation approach choice:

The utils.py module has a `get_output` function which optionally _could_ be used for this, and seemed to work also.
But it prints extra output and it seemed like it would be cleaner to just quietly check for cuda versions and CUDA_HOME
without generating output, as a check that should just quietly pass most of the time -- unless it is to report an error
of course; for that reason I executed the external commands here in place with `subprocess.getoutput` just to quietly
capture the output results I wanted to compare.

dncliss:
Updated this PR -- 
Trying to use python more efficiently, so I've used python regex searching/matching instead
of using pipelines of external commands (grep, awk, cut) for output parsing.

I think this is cleaner, and I've tested various success & failure cases.  So I think these util.py modules are ready for
use (except as of yet they're just not currently being called from anywhere).

NOTE: The version check just returns True/False depending whether the CUDA and cudatoolkit match (doesn't raise
a fatal error).  By contrast, the CUDA_HOME check does raise an error if it doesn't match nvcc's path, as the Issue
requests.  It would be very simple to flip either of these choices (i.e. fatal error versus just a T/F match check), if desired,
or they could be left as coded.

dncliss:
This worked for me locally, but when it gets run in the CI build tests, the version check execution of `nvcc --version` fails because, per the error, the nvcc command is not found.  (?)   (Didn't fail locally, as nvcc is in my path).

I'm surprised that `nvcc` wasn't found in the PATH by default during the CI tests?
I'll see if I can tell why.  (Although not complete due to this error, I'll go ahead and add reviewers.)

bnemanich:
The CI images do not have CUDA installed within them. The tests shouldn't rely on having nvcc installed. You might need to mock the call to nvcc within the tests.

open-ce/build_feedstock.py
```diff
@@ -103,6 +103,8 @@ def build_feedstock_from_command(command, # pylint: disable=too-many-arguments,
     Build a feedstock from a build_command object.
     '''
     utils.check_if_conda_build_exists()
+    utils.check_cuda_version_match()
```
bnemanich, line 106: These should only be called if the `build_type` is cuda.

open-ce/utils.py
```diff
@@ -219,6 +219,74 @@ def cuda_driver_installed():
 
         raise OpenCEError(Error.ERROR, "lsmod command unexpectedly failed") from err
 
+def check_cuda_version_match():
+    '''
```
bnemanich, line 223: This function should be checking that the value of `command.cudatoolkit` is equivalent to the version of CUDA that is installed within the CUDA_HOME environment variable.

open-ce/utils.py
```diff
@@ -219,6 +219,74 @@ def cuda_driver_installed():
...
+    raise OpenCEError(Error.ERROR, "nvcc version does not match system's CUDA version")
+
+def check_cuda_home():
+    '''
+    Ensure CUDA_HOME is set with same PATH as nvcc.
+      If unset, then default it to the PATH of nvcc
+      If paths are different, treat as fatal error.
+    '''
+    try:
+        nvcc_path = subprocess.getoutput("/usr/bin/which nvcc")
```
bnemanich, line 265: This shouldn't be an error. It could be that `nvcc` is installed under `$CUDA_HOME/bin`, but that hasn't been added to the path.

bnemanich: Let's not check whether `nvcc` is in the system path.

open-ce/build_feedstock.py
dncliss: Makes sense; Now only invokes if `cuda` was specified in build type.

open-ce/utils.py
dncliss: `nvcc` is no longer invoked at all. We use $CUDA_HOME if set, or set specifically to default of `/usr/local/cuda` if not previously set. (So, no failure if path doesn't match `nvcc` in system path).

open-ce/utils.py
dncliss: Changed to check version from `command.cudatoolkit` with version within CUDA_HOME (checked via what is reported in the `$CUDA_HOME/version.txt` file).  Now, issue a WARNING if we can't get the version.

open_ce/utils.py
```diff
@@ -226,6 +226,40 @@ def cuda_driver_installed():
...
+
+    # Version is recorded in $CUDA_HOME/version.txt; make sure $CUDA_HOME is set first.
+    # (If $CUDA_HOME doesn't exist, read will fail so we return False).
+    check_cuda_home()
+    cudavers_file = os.environ['CUDA_HOME'] + '/version.txt'
+    try:
+        cudafile = open(cudavers_file,'r')
+        cuda_home_version = cudafile.readline()
+        version_match = cuda_home_version.find(str(command.cudatoolkit))
+        cudafile.close()
```
npanpaliya, line 243: Should this line be put into finally block?

open_ce/build_feedstock.py
```diff
@@ -104,6 +104,9 @@ def build_feedstock_from_command(command, # pylint: disable=too-many-arguments,
     '''
     utils.check_if_conda_build_exists()
 
+    if 'cuda' in command.build_type:
```
npanpaliya, line 107: Adding this check here would make it call for every build command when we are building any open-ce environment. I think this call should be moved to `build_feedstock` entry point function and then added into build_tree's constructor too so that it can be called when `build env ` command is triggered.

open_ce/utils.py
```diff
@@ -226,6 +226,40 @@ def cuda_driver_installed():
...
+        if version_match > 0:
+            return True
+    except Exception:  # pylint: disable=broad-except
+        # Treating as a warning rather than a fatal error
+        print("WARNING: Could not read version from " + cudavers_file)
```
cdeepali, line 248: Do we need to return False even here?

open_ce/build_feedstock.py
```diff
@@ -104,6 +104,9 @@ def build_feedstock_from_command(command, # pylint: disable=too-many-arguments,
...
+        utils.check_cuda_version_match(command)
```
cdeepali, line 108: Do we need to stop build here, with an error in case the version does not match?

cdeepali:
I think we can also add a test for the newly added functionality.

open_ce/utils.py
dncliss: I could change this - do you think that is better?
In this simple usage case, since I was opening the file only to read one line (to read the version) and am then done with
the file, I thought I would close it immediately after the use.  (If there was an open error, the 'except' will
catch that). So I am not sure if it makes much difference in this case if I use a finally or not....  unless that is simply
desirable as better python coding practice  (?)

open_ce/build_feedstock.py
dncliss: That is what I did in this PR originally (I made it a fatal error).  However, this caused the CI build to fail to complete,
because during the CI, three factors combine to make it fail there: (1) CI includes cuda build, (2) CUDA_HOME isn't pre-specified in CI (so this code sets it to default /usr/local/cuda), but (3) the /usr/local/cuda directory isn't present in the CI build -- so if this is a "fatal" error then the CI stops.
Thus, I changed the function to print a WARNING in this situation rather than fatally end the build.  This provides a check and notification that does not exist today (before this PR exists).  Maybe if there's a way to check explicitly that
this is a CI build, I could do that; otherwise making this just a non-fatal warning wouldn't kill the build but would still
serve a notice useful to someone building outside of the CI (if they are watching for warnings).

open_ce/utils.py
npanpaliya: Yes, I think it is always better to release resources in the finally block as it will be always executed. But in your case, if it is just a single line being read from a file during which there is hardly a chance of any mishap, you can skip this comment. As a better python practice, indeed it should go to finally block.

dncliss: Thanks. I'm thinking: even if it seems not to be significant in this case, since I am not an expert with the python language I should be as strict as I can about following best practices, so I will change this as per this advice.

dncliss: @npanpaliya After making the suggested change, the code bombs if the file open() had failed (i.e. if the file wasn't there). If the open fails, the exception message is printed and then the 'finally' with the close() is attempted, but this gives an error since the file isn't open (doesn't exist); we get an `UnboundLocalError` at that line.
In the previous form, the close is only attempted if we are still in the 'try' block at that point, which means the open had succeeded.   So it seems that that I should put this back how it was.

open_ce/build_feedstock.py
dncliss: @npanpaliya I moved this check into the `build_feedstock` entry point function as suggested, just before it calls `build_feedstock_from_command`. In reviewing the constructor in `build_tree.py` I am trying to follow the thinking of adding the same check in there, and where it would be appropriate to do that. (I think I only partially understand this so I will review it some more.)

open_ce/build_feedstock.py
cdeepali: I see.  Yes, so I think for the case when  `/usr/local/cuda/version.txt`  isn't present WARNING is good. 

However, when `/usr/local/cuda/version.txt` is present, but versions do not match ,the build will be successful without any message to the user. Do we want to give a WARNING/Error in this case ?

open_ce/utils.py
npanpaliya: @dncliss - This is how I tried and it worked -
```
diff --git a/open_ce/utils.py b/open_ce/utils.py
index 85b3e05..197f01b 100644
--- a/open_ce/utils.py
+++ b/open_ce/utils.py
@@ -236,16 +236,19 @@ def check_cuda_version_match(command):
     # (If $CUDA_HOME doesn't exist, read will fail so we return False).
     check_cuda_home()
     cudavers_file = os.environ['CUDA_HOME'] + '/version.txt'
+    cudafile = None
     try:
         cudafile = open(cudavers_file,'r')
         cuda_home_version = cudafile.readline()
-        cudafile.close()
         version_match = cuda_home_version.find(str(command.cudatoolkit))
         if version_match > 0:
             return True
     except Exception:  # pylint: disable=broad-except
         # Treating as a warning rather than a fatal error
         print("WARNING: Could not read version from " + cudavers_file)
+    finally:
+        if cudafile:
+            cudafile.close()
```

dncliss: @npanpaliya Ah, that makes complete sense; it needs to be initialized and checked so the close() won't be attempted if the open() had failed.  I'll correct and push this part again.

open_ce/build_feedstock.py
dncliss: @cdeepali Yes, that was the intent; an earlier iteration did so as I remember testing it, but I somehow lost this case along the way.  I've now added a warning message specific for the mismatch case -- not here, but in `utils.py` when returning.  (A T/F return value is provided by the utility function for possible use, but is not actually used here.)

dncliss:
This is being closed as it won't ever be merged in this repository; equivalent PR was opened in `open-ce-builder` to cover the same thing.

PR #329, opened 2021-02-23
Packaging tools

### LINKED ISSUE #299: [REQUEST] Package the tools.
It would be much easier to use the tools if they were installable as a package.

If we do a pip package it can be installed directly from the github repo:

`pip install git+https://github.com/open-ce/open-ce.git@open-cev1.1`

We should also look into making a conda package as well.

npanpaliya:

## Description

Fixes #299 .

## Review process to land 


npanpaliya:
I've created this as a Draft PR to get the initial feedback on the implementation. I'm yet to do various things like -
1. Creating build scripts to create pip package
2. Creating conda recipe
3. Updating documentation

With the current implementation, I'm able to do `pip install git+https://github.com/npanpaliya/open-ce.git@packaging-tools`.

npanpaliya:
@bnemanich , @jayfurmanek and @bethune-bryant - Could you please take a glance on this PR and provide me an early feedback on approach?

open_ce/docker_build.py
```diff
@@ -178,7 +177,7 @@ def build_in_container(image_name, args, arg_strings):
     _start_container(container_name)
 
     # Execute build command
-    cmd = "python {} {} {} {}".format(os.path.join(HOME_PATH, "open-ce", "open-ce", "open-ce"),
+    cmd = "source $HOME/.bashrc; python {} {} {} {}".format(os.path.join(HOME_PATH, "open_ce", "open-ce"),
```
bnemanich, line 180: Do you know why this is needed? I would rather not hard code this in, since users might not have a `.bashrc` file.

npanpaliya: Yes. Earlier we used to copy cloned copy of open-ce code into container. But now, we are copying the files from installed package with imports like `import open_ce.utils`. To make this work inside container, PYTHONPATH needed to be exported which I did in Dockerfiles. However, at this point when python build env is invoked inside container, PYTHONPATH wasn't in effect until `source ~/.bashrc` is done.

npanpaliya: We are writing stuff in ~/.bashrc in our Dockerfiles for build images. So, it will always be present.

bnemanich: I see. I forgot this was within the docker_build code, so we have more control over what will be there. I think this is OK then.

open_ce/images/builder-cuda-ppc64le/Dockerfile.cuda-10.2
```diff
@@ -48,4 +48,5 @@ RUN export PATH="${PATH}" && \
     sudo rm -f /run/nologin && \
     yes | ssh-keygen -f $HOME/.ssh/id_rsa -q -t rsa -N '' && cp $HOME/.ssh/id_rsa.pub $HOME/.ssh/authorized_keys && \
     echo ". $CONDA_HOME/etc/profile.d/conda.sh" >> ${HOME}/.bashrc && \
+    echo "export PYTHONPATH=${PYTHONPATH}:$HOME/open_ce" >> ${HOME}/.bashrc && \
```
bethune-bryant, line 51: Would it be better to pip install the package, instead of setting the python path in the bashrc file?

That way it's being used more as intended?

bethune-bryant:
After making the above comment, I realized it probably wouldn't be straightforward to pip install the package, since it will already be running as a pip installed package, then I got to wondering how it's working now. In your changes here, it's copying in the open-ce folder where the docker_build file is, which would copy the libs from the module, but not the `open-ce` script in `bin`.

https://github.com/npanpaliya/open-ce/blob/8bb678d077cb8639ed9c3b9455685bfe1bd9414f/open_ce/docker_build.py#L164..L165

This shouldn't work, because the `open-ce` entry point should be in `anaconda3/envs/<env>/bin/`, not in `anaconda3/envs/<env>/lib/python3.6/site-packages/open_ce/`.

It turns out the `open-ce` script is currently being installed in both locations. So we probably need to fix that in the `setup.py` file.

Also, once that's fixed we'll have to figure out how to execute `open-ce` inside the container, it will probably have to be copied in separately.

I've tried to find a way to pip install the package in the container, because I think that would be the best option, but I can't find a way to do that.

setup.py
```diff
@@ -0,0 +1,72 @@
...
+setup(
+    name="open-ce",
+    version=open_ce.__version__,
+    description="Open-CE tools for building feedstocks",
+    long_description=get_readme(),
+    url="https://github.com/open-ce/open-ce",
+    author="Open-CE Dev Team",
+    author_email="",
+    packages = find_packages(),
+    scripts = ["open_ce/open-ce"],
```
bethune-bryant, line 45: Instead of scripts, we probably want to define this as the enrty point:
```
entry_points={
        'console_scripts': [
            'open-ce = open-ce:main',
        ],
    },
```

Something along those lines.

npanpaliya: I'd tried this earlier. I've also CONSOLE_SCRIPTS defined above in the setup.py (which I forgot to clean). But it didn't work. Earlier it was not putting the `open-ce` into `bin` directory and then it failed to start.

npanpaliya:
> After making the above comment, I realized it probably wouldn't be straightforward to pip install the package, since it will already be running as a pip installed package, then I got to wondering how it's working now. In your changes here, it's copying in the open-ce folder where the docker_build file is, which would copy the libs from the module, but not the `open-ce` script in `bin`.
> 
> https://github.com/npanpaliya/open-ce/blob/8bb678d077cb8639ed9c3b9455685bfe1bd9414f/open_ce/docker_build.py#L164..L165
> 
> This shouldn't work, because the `open-ce` entry point should be in `anaconda3/envs/<env>/bin/`, not in `anaconda3/envs/<env>/lib/python3.6/site-packages/open_ce/`.
> 
> It turns out the `open-ce` script is currently being installed in both locations. So we probably need to fix that in the `setup.py` file.
> 
> Also, once that's fixed we'll have to figure out how to execute `open-ce` inside the container, it will probably have to be copied in separately.
> 
> I've tried to find a way to pip install the package in the container, because I think that would be the best option, but I can't find a way to do that.

Yes, I had also thought on it earlier but didn't find a way. Unless we know from where (git tag/branch/local code copy) we did pip install on host system programmatically inside `docker_build.py` , we can't do a pip install inside container. 

I agree, open-ce is being copied at two locations currently, which I've fixed now. Now, I need to copy `open-ce` binary at the right location inside the container.

open_ce/images/builder-cuda-ppc64le/Dockerfile.cuda-10.2
npanpaliya: We discussed this and decided to keep it as it is as of now.

bnemanich (approved):

setup.py
bethune-bryant: Okay, what you have is working. If we find there is a better way to do it later we can always change it.

tests/README.md
```diff
@@ -15,6 +15,13 @@ $ conda install pytest pytest-cov pytest-mock
 
 ## Running Tests
 
+Before running the tests, install the Open-CE pip package using 
+following command - 
+```shell
+$ python setup.py install
```
bethune-bryant, line 21: Would it be better to recommend calling pip directly?
```suggestion
$ pip install -e .
```

bethune-bryant (approved):
This is awesome, I'm really excited about this!

All it will take to do a build from scratch is:
```
(base) [bnelson@dlw05 ~]$ conda create -n test python=3.7 conda-build
...
(base) [bnelson@dlw05 ~]$ conda activate test
(test) [bnelson@dlw05 ~]$ pip install git+https://github.com/open-ce/open-ce.git@main
...
(test) [bnelson@dlw05 ~]$ open-ce build env --docker_build --build_types cuda https://raw.githubusercontent.com/open-ce/open-ce-environments/main/envs/xgboost-env.yaml
...
```

I can already do this using your branch!

tests/README.md
npanpaliya: Yeah. That will also do! I'll change the documentation.

PR #339, opened 2021-03-02
Git tool updates

bethune-bryant:
Updates to the apply_patches script to add reviewers and add param replacement in the patch.

### DIFF
--- a/git_tools/git_utils.py
+++ b/git_tools/git_utils.py
@@ -228,7 +228,7 @@ def fill_in_params(filename, params=None, **kwargs):
         params = dict()
 
     for key, value in params:
-        text = text.replace("${{{0}}}".format(key), value)
+        text = text.replace("${{{}}}".format(key), value)
 
     for key, value in kwargs.items():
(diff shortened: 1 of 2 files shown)


git_tools/git_utils.py
```diff
@@ -175,3 +215,26 @@ def get_current_branch(repo_path):
 def apply_patch(repo_path, patch_path):
     '''Apply a patch to the given repo.'''
     _execute_git_command(repo_path, "cat \"{}\" | git am -3 -k".format(patch_path))
+
+def fill_in_params(filename, params=None, **kwargs):
```
bnemanich, line 219: This probably could have been done with something like jinja instead of writing this your self. However, since it is for a specific use case maybe that would have been overkill.

bethune-bryant: Yeah, it may get weird putting jinja inside of patches too. If we ever need to have this do more stuff or be more robust, we could switch it over.

bnemanich (approved):

jayfurmanek (approved):

Issue #325, opened 2021-02-19
[BUG] build image gives error shutil.SameFileError


npanpaliya:
**Describe the bug**
`open-ce/open-ce/open-ce build image` command gives error `shutil.SameFileError`.

**To Reproduce**
Run the following command - 
open-ce/open-ce/open-ce build image --conda_env_file=condabuild/opence-conda-env-py3.7-cuda-openmpi-11.0.yaml --local_conda_channel=./condabuild

**Expected behavior**
The command should be completed successfully.

Earlier, conda env file used to get generated at the location from where `build env` is run. So, build image was implemented in such a way that it used to copy the conda env file from the original location to the directory specified as `local_conda_channel` to modify it as per docker container paths. But later on, location of conda env file being generated by `build env` is changed to be in the `output_folder` itself which is nothing but `local_conda_channel`. And hence https://github.com/open-ce/open-ce/blob/master/open-ce/build_image.py#L84 throws error.

**Execution Logs**
If applicable, add logs to help explain your problem.

**System (please complete the following information):**
 - OS: [e.g. RHEL, Windows]
 - Architecture [e.g. Power, x86]

**Additional context**

PR #347, opened 2021-03-05
Enable cuda in runtime image

jayfurmanek:

- [x] Did you update any affected [documentation](https://github.com/open-ce/open-ce/blob/main/doc/)?
- [x] Did you write any [tests](https://github.com/open-ce/open-ce/blob/main/tests/) to validate this change?  

## Description

This PR enables GPUs in the runtime DockerFile when using nvidia-container-runtime.
This also makes the conda env file consumption optional and more robust.

## Review process to land 

### DIFF
--- a/open_ce/build_image.py
+++ b/open_ce/build_image.py
@@ -84,27 +84,27 @@ def build_runtime_container_image(args):
         conda_env_file = os.path.abspath(conda_env_file)
         _validate_input_paths(local_conda_channel, conda_env_file)
 
-    # Copy the conda environment file into the local conda channel with a new name and modify it
-    conda_env_runtime_filename = os.path.splitext(os.path.basename(args.conda_env_file))[0]+'-runtime.yaml'
-    conda_env_runtime_file = os.path.join(local_conda_channel, os.path.basename(conda_env_runtime_filename))
-    try:
-        shutil.copy(conda_env_file, conda_env_runtime_file)
-    except shutil.SameFileError:
-        print("Info: Environment file already in local conda channel.")
-    utils.replace_conda_env_channels(conda_env_runtime_file, r'file:.*', "file:/{}".format(TARGET_DIR))
+        # Copy the conda environment file into the local conda channel with a new name and modify it
+        conda_env_runtime_filename = os.path.splitext(os.path.basename(args.conda_env_file))[0]+'-runtime.yaml'
+        conda_env_runtime_file = os.path.join(local_conda_channel, os.path.basename(conda_env_runtime_filename))
+        try:
+            shutil.copy(conda_env_file, conda_env_runtime_file)
+        except shutil.SameFileError:
+            print("Info: Environment file already in local conda channel.")
+        utils.replace_conda_env_channels(conda_env_runtime_file, r'file:.*', "file:/{}".format(TARGET_DIR))
 
         # Check if input local conda channel path is absolute
         if os.path.isabs(args.local_conda_channel):
             # make it relative to BUILD CONTEXT
             args.local_conda_channel = os.path.relpath(args.local_conda_channel, start=BUILD_CONTEXT)
 
-    image_name = build_image(args.local_conda_channel, os.path.basename(conda_env_runtime_file))
+        image_name = build_image(args.local_conda_channel, os.path.basename(conda_env_runtime_file))
 
-    # Remove the copied environment file
-    try:
-        os.remove(conda_env_runtime_file)
-    except OSError:
-        print("Info: Temporary environment file cannot be removed.")
+        # Remove the copied environment file
+        try:
+            os.remove(conda_env_runtime_file)
+        except OSError:
+            print("Info: Temporary environment file cannot be removed.")
 
         print("Docker image with name {} is built successfully.".format(image_name))
 
--- a/open_ce/build_image.py
+++ b/open_ce/build_image.py
@@ -100,7 +100,7 @@ def build_runtime_docker_image(args):
     try:
         os.remove(conda_env_runtime_file)
     except OSError:
-        print("Info: Temporary environment cannot be removed.")
+        print("Info: Temporary environment file cannot be removed.")
 
     print("Docker image with name {} is built successfully.".format(image_name))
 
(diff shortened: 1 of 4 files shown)


open-ce/build_image.py
```diff
@@ -81,7 +81,10 @@ def build_runtime_docker_image(args):
     _validate_input_paths(local_conda_channel, conda_env_file)
 
     # Copy the conda environment file into the local conda channel to modify it
-    shutil.copy(conda_env_file, local_conda_channel)
+    try:
+        shutil.copy(conda_env_file, local_conda_channel)
```
npanpaliya, line 85: This copying was earlier added because conda env file location was different than the output_folder. 
We do want to copy but not to local_conda_channel as it will already have it. And copy is also must because we modify channels in conda env file according to runtime docker container, which would make the env file erroneous if someone uses it to create a conda environment on the host using it.

jayfurmanek: yeah, I considered making a back up in this case, but there are some things to consider: 
 - We could move the original to SOMEWHERE/FILE.orig, but then if the user runs again, it will just be overwritten by the back up of the edited file.
 - So we could move the original to SOMEWHERE/FILE.orig.date (or something) but then we are accruing a bunch of garbage along side the one original as a new backup is made on each run. That's annoying an unhelpful.
 
It's also a bit unfortunate because it's very likely the env file that is being used will be located in the conda channel already because we put it there after the build. The behavior as it is now fails out on this SAMEFILE error and ends, so it must be handled. I decided to just not copy and this case which does lose the original, but..I think it's not so bad. The user can back it up (or manually edit it) if they need to go back and use the env file manually.

One other comment - we don't have or use a message logger in OpenCE yet (perhaps we should), so I'm just manually doing a print here.

npanpaliya: I think we can just create the copy of the env file in the same output_folder (local_conda_channel) with name like `opence-conda-env-py3.6-cpu-openmpi-10.2-docker-runtime.yaml` and edit that and pass it onto the docker image. This way, the original file will remain intact for user to work on host.

jayfurmanek: yeah - that's doable.

npanpaliya: Yeah, after we are done with the image creation, we can delete this new copy. That way host won't have files that are useless for it.

open-ce/build_image.py
```diff
@@ -80,17 +80,27 @@ def build_runtime_docker_image(args):
...
         args.local_conda_channel = os.path.relpath(args.local_conda_channel, start=BUILD_CONTEXT)
 
-    image_name = build_image(args.local_conda_channel, os.path.basename(conda_env_file))
+    image_name = build_image(args.local_conda_channel, os.path.basename(conda_env_runtime_file))
+
+    # Remove the copied environment file
+    try:
+        os.remove(conda_env_runtime_file)
+    except OSError:
+        print("Info: Temporary environment cannot be removed.")
```
npanpaliya, line 103: Should it be Temporary environment file?

tests/build_image_test.py
```diff
@@ -127,29 +120,7 @@ def test_channel_update_in_conda_env(mocker):
         side_effect=(lambda x: helpers.validate_cli(x, expect=["docker build",
                                                                "-t " + intended_image_name])))
 
-    channel_index_before, _ = get_channel_being_modified("tests/test-conda-env.yaml")
```
npanpaliya, line 130: I thought it was good to have this test to ensure conda env file is modified before build image. Any reason for removing this?

npanpaliya:
Fixes this issue https://github.com/open-ce/open-ce/issues/325

tests/build_image_test.py
jayfurmanek: We are removing the edited temporary environment file now and this test only worked assuming it remained there after the fact.

open-ce/build_image.py
jayfurmanek: yup!

tests/build_image_test.py
bnemanich: You could mock the os.remove function to not do anything. That way we could still verify that the modifications are working properly.

tests/build_image_test.py
```diff
@@ -176,6 +145,4 @@ def test_build_image_with_container_build_args(mocker):
 
     build_args = "--build-arg ENV1=test1 --build-arg ENV2=test2 same_setting 0,1"
     arg_strings = ["build", build_image.COMMAND, "--local_conda_channel", "tests/testcondabuild", "--conda_env_file", "tests/test-conda-env.yaml", "--container_tool", "podman", "--container_build_args", build_args]
-    opence._main(arg_strings)
-
-    os.remove("tests/testcondabuild/test-conda-env.yaml")
+    opence._main(arg_strings)
```
bnemanich, line 148: There is a newline missing here.

open_ce/build_image.py
```diff
@@ -84,10 +84,14 @@ def build_runtime_container_image(args):
         conda_env_file = os.path.abspath(conda_env_file)
         _validate_input_paths(local_conda_channel, conda_env_file)
 
-        # Copy the conda environment file into the local conda channel to modify it
-        shutil.copy(conda_env_file, local_conda_channel)
-        conda_env_file = os.path.join(local_conda_channel, os.path.basename(conda_env_file))
-        utils.replace_conda_env_channels(conda_env_file, r'file:.*', "file:/{}".format(TARGET_DIR))
+        # Copy the conda environment file into the local conda channel with a new name and modify it
+        conda_env_runtime_filename = os.path.splitext(os.path.basename(args.conda_env_file))[0]+'-runtime.yaml'
```
bethune-bryant, line 88: This should be args.conda_env_files
```suggestion
        conda_env_runtime_filename = os.path.splitext(os.path.basename(args.conda_env_files))[0]+'-runtime.yaml'
```

bethune-bryant (changes requested):

open_ce/build_image.py
jayfurmanek: yeah thanks. I had to merge an update in this one yet again and I used the web merge this time while on the phone. Not recommended. :) I'll fix those.

jayfurmanek: actually this should just be
```
conda_env_runtime_filename = os.path.splitext(os.path.basename(conda_env_files))[0]+'-runtime.yaml'
```
since we are working with one of the env files in the list at this point.

tests/build_image_test.py
jayfurmanek: Good idea. I added that back with the mock.

bethune-bryant (approved):

PR #351, opened 2021-03-08
Updated pylint to 2.7 and fixed corresponding new errors

### LINKED ISSUE #341: Update pylint version
Update the version of pylint that is used during PR tests to 2.7.x.

Fix any issues that may arise with the newer version of pylint.

npanpaliya:

## Description

Fixes #341.

After upgrading pylint to 2.7.x, most of the pylint errors are `invalid-name` pointing to the Enum keys we have defined in env_config.py, test_feedstock.py or get_licenses.py. I think changing all those keys everywhere will also need us change the environment files. So, I think we should add this error to the exception list. My PR for #329 has already added it in exception list (disabled in .pylintrc) as it was complaining for `invalid name` of `open-ce` binary file.

## Review process to land 

### DIFF
--- a/.github/workflows/builder-unit-test.yml
+++ b/.github/workflows/builder-unit-test.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Install dependencies
         shell: bash -l {0}
         run: |
-          conda install -y pylint=2.6.0 \
+          conda install -y pylint=2.7 \
                            conda-build \
                            pytest=6.0.1 \
                            pytest-cov=2.10.1 \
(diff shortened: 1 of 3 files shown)


cdeepali (approved):

bnemanich (approved):

jayfurmanek (approved):

.github/workflows/builder-unit-test.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       - name: Install dependencies
         shell: bash -l {0}
         run: |
-          conda install -y pylint=2.6.0 \
+          conda install -y pylint=2.7 \
```
bethune-bryant, line 35: We'll need to keep an eye on issues that come up with specifying `2.7` instead of a specific `2.7.X`.

PR #355, opened 2021-03-08
Add Indirect Dependency Checking

### LINKED ISSUE #345: [BUG] DAG Generation When Dependency Chain Includes External Dependencies
**Describe the bug**

image

In the above dependency graph, let's say:

- A, B, C, E and F are all internal packages being built by `open-ce` 
- D is an external dependency that has a dependency on the internal package F

This would cause a race condition since the Open-CE tools won't know about the dependency of A and C on F.
The DAG generated by `open-ce` would show that no packages depended on F.

**Expected behavior**
We should look into adding external dependency traversing to find any later dependencies on internal packages.

bethune-bryant:

## Description

Fixes #345

### DIFF
--- a/open_ce/build_env.py
+++ b/open_ce/build_env.py
@@ -122,17 +122,6 @@ def build_env(args):
                                conda_build_config=args.conda_build_config,
                                packages=inputs.parse_arg_list(args.packages))
 
-    #import matplotlib.pyplot as plt
-    #import networkx
-    #plt.figure(figsize=(200,200))
-    #networkx.draw_circular(build_tree._tree, with_labels=True)
-    #plt.savefig(os.path.join(os.path.abspath(args.output_folder), 'output.png'))
-    #print("Cycles: ", list(networkx.simple_cycles(build_tree._tree)))
-    #print("Build order:")
-    #print(list(build_tree))
-    #for x in self:
-    #    print("Build", x)
-
     # Generate conda environment files
     conda_env_files = build_tree.write_conda_env_files(output_folder=os.path.abspath(args.output_folder),
                                                        path=os.path.abspath(args.output_folder))
--- a/open_ce/build_env.py
+++ b/open_ce/build_env.py
@@ -122,6 +122,17 @@ def build_env(args):
                                conda_build_config=args.conda_build_config,
                                packages=inputs.parse_arg_list(args.packages))
 
+    #import matplotlib.pyplot as plt
+    #import networkx
+    #plt.figure(figsize=(200,200))
+    #networkx.draw_circular(build_tree._tree, with_labels=True)
+    #plt.savefig(os.path.join(os.path.abspath(args.output_folder), 'output.png'))
+    #print("Cycles: ", list(networkx.simple_cycles(build_tree._tree)))
+    #print("Build order:")
+    #print(list(build_tree))
+    #for x in self:
+    #    print("Build", x)
+
     # Generate conda environment files
     conda_env_files = build_tree.write_conda_env_files(output_folder=os.path.abspath(args.output_folder),
                                                        path=os.path.abspath(args.output_folder))
--- a/open_ce/build_env.py
+++ b/open_ce/build_env.py
@@ -48,6 +48,7 @@ ARGUMENTS = [Argument.CONDA_BUILD_CONFIG,
              Argument.CONTAINER_BUILD,
              Argument.GIT_LOCATION,
              Argument.GIT_TAG_FOR_ENV,
+             Argument.GIT_UP_TO_DATE,
              Argument.TEST_LABELS,
              Argument.CONTAINER_BUILD_ARGS,
              Argument.CONTAINER_TOOL]
@@ -119,6 +120,7 @@ def build_env(args):
                                channels=args.channels_list,
                                git_location=args.git_location,
                                git_tag_for_env=args.git_tag_for_env,
+                               git_up_to_date=args.git_up_to_date,
                                conda_build_config=args.conda_build_config,
                                packages=inputs.parse_arg_list(args.packages))
 
--- a/open_ce/build_tree.py
+++ b/open_ce/build_tree.py
@@ -269,6 +269,7 @@ class BuildTree(): #pylint: disable=too-many-instance-attributes
                  channels=None,
                  git_location=utils.DEFAULT_GIT_LOCATION,
                  git_tag_for_env=utils.DEFAULT_GIT_TAG,
+                 git_up_to_date=False,
                  conda_build_config=utils.DEFAULT_CONDA_BUILD_CONFIG,
                  packages=None):
 
@@ -277,6 +278,7 @@ class BuildTree(): #pylint: disable=too-many-instance-attributes
         self._channels = channels if channels else []
         self._git_location = git_location
         self._git_tag_for_env = git_tag_for_env
+        self._git_up_to_date = git_up_to_date
         self._conda_build_config = conda_build_config
         self._external_dependencies = dict()
         self._conda_env_files = dict()
@@ -428,6 +430,7 @@ class BuildTree(): #pylint: disable=too-many-instance-attributes
         # at all specified then fall back to default branch of the repo.
 
         git_tag = self._git_tag_for_env
+        git_tag_for_package = None
         if git_tag is None:
             git_tag_for_package = package.get(env_config.Key.git_tag.name, None) if package else None
             if git_tag_for_package:
@@ -435,7 +438,7 @@ class BuildTree(): #pylint: disable=too-many-instance-attributes
             else:
                 git_tag = env_config_data.get(env_config.Key.git_tag_for_env.name, None) if env_config_data else None
 
-        clone_successful = utils.git_clone(git_url, git_tag, repo_dir)
+        clone_successful = utils.git_clone(git_url, git_tag, repo_dir, self._git_up_to_date and not git_tag_for_package)
 
         if clone_successful:
             patches = package.get(env_config.Key.patches.name, []) if package else []
--- a/open_ce/build_tree.py
+++ b/open_ce/build_tree.py
@@ -309,7 +309,8 @@ class BuildTree(): #pylint: disable=too-many-instance-attributes
                 for package in packages:
                     if not {n for n in traverse_build_commands(variant_tree, return_node=True) if package in n.packages}:
                         print("INFO: No recipes were found for " + package + " for variant " + variant_string)
-                variant_start_nodes = {n for n in traverse_build_commands(variant_tree, return_node=True) if n.packages.intersection(packages)}
+                variant_start_nodes = {n for n in traverse_build_commands(variant_tree, return_node=True)
+                                            if n.packages.intersection(packages)}
 
             self._initial_nodes += variant_start_nodes
 
--- a/open_ce/build_tree.py
+++ b/open_ce/build_tree.py
@@ -17,6 +17,7 @@
 """
 
 import os
+import networkx
 from open_ce import utils
 from open_ce import env_config
 from open_ce import validate_config
@@ -95,6 +96,13 @@ class BuildCommand():
 
         return build_args
 
+    def all_outputs_exist(self, output_folder):
+        """
+        Returns true if all of the output_files already exist.
+        """
+        return all((os.path.exists(os.path.join(os.path.abspath(output_folder), package))
+                    for package in self.output_files))
+
     def name(self):
         """
         Returns a name representing the Build Command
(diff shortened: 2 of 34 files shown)


bethune-bryant:
I now have this working for the libprotobuf case.

If I remove `libprotobuf` from the protobuf recipe, then the current version of the tools fail validation, but with these changes validation is successful. This is the command:
open-ce validate config --python_versions 3.8 --build_types cuda --repository_folder /tmp/bryant https://raw.githubusercontent.com/open-ce/open-ce-environments/main/envs/tensorflow-env.yaml

open_ce/build_tree.py
```diff
@@ -108,93 +114,96 @@ def name(self):
...
-    yield from _traverse_build_commands(commands, is_seen, deps)
-
-def _traverse_build_commands(commands, is_seen, deps):
-    for dep in deps:
-        if is_seen[dep]:
-            continue
-        yield from _traverse_build_commands(commands, is_seen, commands[dep].build_command_dependencies)
-        yield commands[dep]
-        is_seen[dep] = True
+    #pylint: disable=too-many-instance-attributes,too-many-arguments,too-many-locals
```
bnemanich, line 144: I don't think all of these pylint exceptions are needed here.

open_ce/build_tree.py
```diff
@@ -108,93 +114,96 @@ def name(self):
...
+    """
+    This function gets all run dependencies of a node that don't depend on
+    any internal build commands.
+    """
+    deps = set()
+    if node.build_command is not None:
+        run_deps = {x for x in node.build_command.run_dependencies
+                                if utils.remove_version(x) not in map(utils.remove_version, node.packages)}
+        for run_dep in run_deps:
+            run_dep_node = {x for x in tree.successors(node)
```
bnemanich, line 198: What is the pop for?

open_ce/build_tree.py
```diff
@@ -108,93 +114,96 @@ def name(self):
...
+                else:
+                    yield current.build_command
+
+def get_independent_runtime_deps(tree, node):
+    """
+    This function gets all run dependencies of a node that don't depend on
+    any internal build commands.
+    """
+    deps = set()
+    if node.build_command is not None:
```
bnemanich, line 194: Can this be: `if not node.build_command` or do we need to handle the case of `node.build_command == []`

open_ce/build_tree.py
```diff
@@ -108,93 +114,96 @@ def name(self):
...
+        run_deps = {x for x in node.build_command.run_dependencies
+                                if utils.remove_version(x) not in map(utils.remove_version, node.packages)}
+        for run_dep in run_deps:
+            run_dep_node = {x for x in tree.successors(node)
+                                    if utils.remove_version(run_dep) in map(utils.remove_version, x.packages)}.pop()
+            if len({x for x in networkx.descendants(tree, run_dep_node) if x.build_command is not None}) == 0:
```
bnemanich, line 200: Instead of checking the length of the result, you can just say: `if not {...}:`

open_ce/build_tree.py
```diff
@@ -322,52 +282,49 @@ def __init__(self,
...
 
             # If the packages argument is provided, find the indices into the build_commands for all
             # of the packages that were requested.
-            variant_package_indices = []
             if packages:
                 for package in packages:
-                    if package in package_indices:
-                        variant_package_indices += package_indices[package]
-                    else:
+                    if len({n for n in variant_start_nodes if package in n.packages}) == 0:
```
bnemanich, line 311: Can remove length check here as well.

open_ce/build_tree.py
```diff
@@ -396,44 +353,70 @@ def _get_repo(self, env_config_data, package):
...
+                    if package_name in seen:
+                        continue
+                    seen.add(package_name)
+                    package_info = conda_utils.get_latest_package_info(self._channels, package)
+                    dep_graph.add_node(DependencyNode({package}))
+                    for dep in package_info['depends']:
+                        dep_name = utils.remove_version(dep)
+                        local_dest = {dest_node for dest_node in dep_graph.nodes()
+                                                if dep_name in map(utils.remove_version, dest_node.packages)}
+                        if len(local_dest) > 0:
```
bnemanich, line 411: `if local_dest:`

open_ce/build_tree.py
```diff
@@ -513,45 +496,87 @@ def get_test_feedstocks(self, variant_string):
...
-        return [current_branch]
-    result = []
-    for dependency in tree[current]:
-        next_step = find_all_cycles(tree, dependency, current_branch)
-        if next_step:
-            result += next_step
-    return result
-
-def get_installable_packages(build_commands, external_deps, package_indices=None):
+    def _detect_cycle(self):
```
bnemanich, line 499: I don't think we are calling `_detect_cycle` anymore.

open_ce/build_tree.py
```diff
@@ -513,45 +496,87 @@ def get_test_feedstocks(self, variant_string):
...
+        cycle_print = ""
+        try:
+            cycle_print = str(networkx.find_cycle(self._tree, orientation="original"))
+        except networkx.NetworkXNoCycle:
+            return
+        raise OpenCEError(Error.BUILD_TREE_CYCLE, cycle_print)
+
+def _create_edges(tree):
+    for node in set(tree.nodes()):
```
bnemanich, line 508: Why is this converted to a set? Can there be duplicate nodes in a tree?

open_ce/build_tree.py
```diff
@@ -513,45 +496,87 @@ def get_test_feedstocks(self, variant_string):
...
+        if node.build_command is not None:
+            for dependency in node.build_command.get_all_dependencies():
+                local_dest = {dest_node for dest_node in tree.nodes()
+                                        if utils.remove_version(dependency)
+                                            in map(utils.remove_version, dest_node.packages)}
+                if len(local_dest) > 0:
```
bnemanich, line 514: `if local_dest:`

open_ce/build_tree.py
bethune-bryant: You're right. These got copy and pasted.

open_ce/build_tree.py
bethune-bryant: Since build command is an object, I was being explicit with the `not None`. I wasn't sure how python converts objects to bools.

open_ce/conda_utils.py
```diff
@@ -72,3 +75,27 @@ def get_output_file_paths(meta, variants):
...
+    ret_code, std_out, _ = run_command_capture(cli, stderr=subprocess.STDOUT)
+    if not ret_code:
+        raise OpenCEError(Error.CONDA_PACKAGE_INFO, cli, std_out)
+    return std_out
+
+def get_latest_package_info(channels, package):
+    '''
+    Get the conda package info for the most recent search result.
+    '''
+    results = yaml.safe_load(conda_package_info(channels, package))
```
bnemanich, line 96: Doesn't `conda_package_info` return output from a JSON file? If so, should probably use `json.loads`.

open_ce/build_tree.py
bethune-bryant: I'm using the set builder notation to do a filter and since that returns a list I pop off the first element. I could probably do a find.
I looked and I could probably do this a little cleaner with the `next` keyword instead of pop.

open_ce/build_tree.py
bethune-bryant: Yeah, I think so.

open_ce/conda_utils.py
```diff
@@ -72,3 +75,27 @@ def get_output_file_paths(meta, variants):
...
+    retval = results[list(results.keys())[0]][0]
```
bnemanich, line 97: Can probably store the result of `results[list(results.keys())[0]]` since it is used twice and kind of hard to read.

open_ce/build_tree.py
bethune-bryant: Yep!

open_ce/build_tree.py
bethune-bryant: We're not, but I didn't want to get rid of it in case we wanted to use it in the future.

open_ce/build_tree.py
bethune-bryant: It's creating a copy so that it doesn't change during the for loop.

tests/build_tree_test.py
```diff
@@ -345,206 +346,217 @@ def test_check_recipe_path_package_field():
...
-    for build_command in mock_build_tree:
-        output += ' '.join([mock_build_tree[dep].name() for dep in build_command.build_command_dependencies]) + "\n"
-
-    expected_output = "\nrecipe2-py2-6-cpu-openmpi-10-2\nrecipe2-py2-6-cpu-openmpi-10-2 recipe3\n"
-
-    assert output == expected_output
+def sample_build_commands() :
+    retval = networkx.DiGraph()
+    node1 = build_tree.DependencyNode(packages=["package1a", "package1b"], build_command=build_tree.BuildCommand("recipe1",
+                                                                                                                "repo1",
```
bnemanich, line 352: Indentation looks a little off here.

open_ce/conda_utils.py
bethune-bryant: I saw a recommendation somewhere to use yaml safe load even for json. Let me see if I can find the reasoning.

tests/build_tree_test.py
```diff
@@ -345,206 +346,217 @@ def test_check_recipe_path_package_field():
...
-
-    out_commands, _ = build_tree._add_build_command_dependencies(additional_build_commands,initial_build_commands,len(initial_build_commands))
-    assert len(out_commands)==1  # Make sure the non-duplicates are not removed
-
-    out_commands, _ = build_tree._add_build_command_dependencies(duplicate_build_commands, initial_build_commands, len(initial_build_commands))
-    assert len(out_commands)==1
-
-    for build_command in out_commands:
-        assert build_command.build_command_dependencies == [0]
+# def test_build_tree_cycle_fail():
```
bnemanich, line 402: Can we still leave in some sort of test that makes sure it handles cycles correctly? Or maybe that is already happening somewhere else since python introduces cycles.

bethune-bryant: Yeah, that's a good idea. I'll add a test along those lines.

open_ce/conda_utils.py
bethune-bryant: That makes sense.

open_ce/build_tree.py
bnemanich: OK, then we will need to leave a test in a test for it as well.

open_ce/build_tree.py
bnemanich: Ah, can you add a comment here mentioning that?

open_ce/conda_utils.py
bnemanich: Ah OK. I'm guessing due to the "safe" part.

open_ce/build_tree.py
bethune-bryant: That makes sense, so it doesn't decay. I'll do this as well.

open_ce/build_tree.py
bethune-bryant: Yep, done.

open_ce/build_tree.py
bnemanich: I think the "pythonic" way is to just do `if node.build_command:`

open_ce/conda_utils.py
bethune-bryant: I've switched this to json load to be more precise.

open_ce/build_tree.py
bethune-bryant: I changed this to use next.

open_ce/build_tree.py
bethune-bryant: Okay, I'll switch to this.

tests/build_tree_test.py
bethune-bryant: I added cycle detection back, but it only checks packages being built.

open_ce/build_tree.py
bethune-bryant: I added cycle detection back, but it only checks packages being built.

open_ce/build_env.py
```diff
@@ -122,6 +122,17 @@ def build_env(args):
                                conda_build_config=args.conda_build_config,
                                packages=inputs.parse_arg_list(args.packages))
 
+    #import matplotlib.pyplot as plt
```
bnemanich, line 125: Should this code be removed? Or added as some sort of debugging function?

bethune-bryant: Thanks!
I think we might be able to add it as some rendering function in the future.

jayfurmanek (approved):

bnemanich (approved):

PR #365, opened 2021-03-15
Conda recipe

### LINKED ISSUE #349: [REQUEST] Create conda recipe of Open-CE tools
Create a conda recipe of open-CE tools. #329 has added pip package creation mechanism.
We also need scripts to create pip and conda packages of open-ce tools.

npanpaliya:

## Description

Fixes #349 .

Have created conda recipe along with a basic script to create conda package of open-ce tools.
I think the script can be further enhanced to add a functionality of uploading the package to our channel or may be anaconda's channel (not sure about this), every time we tag the release. I've not done that here. 
Please let me know if we want to handle this in this script (if yes, we will need anaconda's login, etc.. which I'm yet to figure out completely) or it should go to git_tools/create_opence_release.py.

## Review process to land 


bethune-bryant (changes requested):
Can we build/install and maybe run a simple test, even something like `open-ce -h`?

Also, does it make sense, (and is it possible) to add a build number check as an optional test to this repo as well, so that we remember to update the build number.

npanpaliya:
@bethune-bryant - I re-read your comment. Please note that I've not created a separate feedstock repo for this. Recipe is pushed  into the same open-ce repo. Do you suggest to create new repo?

bethune-bryant:
> 
> 
> @bethune-bryant - I re-read your comment. Please note that I've not created a separate feedstock repo for this. Recipe is pushed into the same open-ce repo. Do you suggest to create new repo?

No, I don't think we need a separate feedstock, if we can't easily re-use the build number check on this repo we can save that for a future issue.

.github/workflows/builder-unit-test.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         run: |
           pip install -e .
           conda install -y pyyaml requests
-          ./open_ce/open-ce -h
+          open-ce -h
```
bethune-bryant, line 31: Good catch here!

.github/actions/opence-conda-package-required-pr/action.yml
```diff
@@ -0,0 +1,16 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Install dependencies
+      shell: bash -l {0}
+      run: |
+        open_ce_path="$GITHUB_ACTION_PATH/../../../"
+        sh ${open_ce_path}/packaging_scripts/create_conda_package.sh
+        conda install -y open-ce -c file://${open_ce_path}/open-ce-conda-pkg
+        conda list open-ce
```
bethune-bryant, line 16: We probably shouldn't make an action for this, since it's only going to be used in this repo. Instead, you can just add this run step directly to the workflow yaml file.

bethune-bryant (changes requested):

.github/actions/opence-conda-package-required-pr/action.yml
npanpaliya: Yeah, earlier I'd added this in the workflow itself. But it was failing to access open-ce root level directory for some reason. I'll retry that change.

npanpaliya: Adding it back to workflow and fixing the paths worked. Thanks @bethune-bryant for your review and inputs.

.github/workflows/conda-recipe-test.yml
```diff
@@ -0,0 +1,41 @@
...
+          auto-update-conda: true
+          python-version: "3.7"
+      - name: Build and install the conda package
+        id: build_install
+        run: |
+          conda install -y conda-build
+          sh ./packaging_scripts/create_conda_package.sh
+          conda install -y open-ce -c file://$(pwd)/open-ce-conda-pkg
+          conda list open-ce
+
```
bethune-bryant, line 25: We should execute open-ce to make sure things are working as a conda package.
```suggestion
          conda list open-ce
          open-ce build env -h

```

.github/actions/opence-conda-package-required-pr/action.yml
bethune-bryant: Looks good, thanks!

.github/workflows/conda-recipe-test.yml
bethune-bryant: This is the last thing blocking this PR. I'm going to go ahead and do the commit, and approve the PR so that you can move forward.

bethune-bryant:
I added an `open-ce build env` command to the conda test, and it failed.

Something may not be setup right with the conda package. Is there something extra we need to do for the executable part?

npanpaliya:
> I added an `open-ce build env -h` command to the conda test, and it failed.
> 
> The list worked, but the execution failed:
> 
> ```
> # packages in environment at /usr/share/miniconda:
> #
> # Name                    Version                   Build  Channel
> open-ce                   1.1.3                      py_1    file:///home/runner/work/open-ce/open-ce/open-ce-conda-pkg
> /home/runner/work/_temp/8ad1ed06-63ba-463c-af58-236017df8144.sh: line 5: open-ce: command not found
> ```
> 
> Something may not be setup right with the conda package. Is there something extra we need to do for the executable part?

I've updated the workflow to create a new conda environment to install the conda package of open-ce. And it worked fine. Shall I merge this now?

conda/recipe/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  - git_url: ../..
+
+build:
+  number: 1
```
bnemanich, line 12: I'm not sure we want to explicitly add a build number here. I don't want to have to update the build number every time we make a change to the open-ce repo. Maybe there is something else we can use like the GitHub PR number.

conda/recipe/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+  noarch: python
+
+requirements:
+  build:
+    - python
+    - setuptools
+  run:
+    - pyyaml
+    - requests
+    - python
```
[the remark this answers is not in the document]
npanpaliya: conda-build isn't added for pip package too, as we don't expect it to be present when using docker_build.

conda/recipe/meta.yaml
npanpaliya: In that case we would even want to remove github workflow tests that should execute `check_build_numbers.py` on this recipe.

conda/recipe/meta.yaml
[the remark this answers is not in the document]
bnemanich: Yeah, but the reasoning for that is we don't want people to have to install conda. In this case, conda is already installed, so installing `conda-build` won't be any more difficult than installing other packages.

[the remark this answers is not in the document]
npanpaliya: Okay. I'll add at both the places then.

conda/recipe/meta.yaml
npanpaliya: @bnemanich - For build number I've used `GIT_DESCRIBE_NUMBER` which is number of commits since the last tag created on the main branch. So, after every commit/PR merge to main branch it will keep on incrementing until we create a tag. Once the new tag is created, it will be back to 0 and again start incrementing. I think this is the better option than using Github PR number (which can be parsed from $GITHUB_REF) as this variable is empty if we locally try to use it.

There is another side problem that I'm seeing with `git describe --tags` on our open-ce main branch. It shows `open-ce-v1.1.2-38-g6f5c26b` whereas it should have been `open-ce-v1.1.3-35-...`. It could be because 1.1.3 is not tagged from main branch. (_38 is the commit count since 1.1.2 and 35 since 1.1.3 on the main branch_)

setup.py
```diff
@@ -26,6 +26,7 @@ def get_readme():
 
 
 REQUIRED_PACKAGES = [
+    "conda-build",
```
bethune-bryant, line 29: Does this requirement check installed conda packages?

npanpaliya: No, with pip, packages listed here would be checked for pip packages instead of conda one. So, if conda-build is already installed via conda, pip might not detect it and install pip package of conda-build too. I've not tried this although.

bnemanich: Yeah, it feels like we shouldn't add this here. I'm not sure what advantage it gives us, and might cause some errors.

npanpaliya: Removed conda-build.

bnemanich (approved):

bethune-bryant (approved):

Issue #326, opened 2021-02-22
[REQUEST] Try building Tensorflow with mkl support (oneAPI variant)


npanpaliya:
Currently MKL is disabled for TF. We need to investigate if it can be enabled now.

jayfurmanek:
@npanpaliya Please push up your changes to build the mkl version.

PR #367, opened 2021-03-18
Set MKL_FEATURE to OFF if it is unset

npanpaliya:

## Description

Fixes https://github.com/open-ce/open-ce/issues/326.

## Review process to land 

### DIFF
--- a/open_ce/utils.py
+++ b/open_ce/utils.py
@@ -302,5 +302,5 @@ def detect_mkl_feature():
     If it is not set at all, the method sets it to OFF.
     '''
     if not 'MKL_FEATURE' in os.environ:
-        print("MKL_FEATURE isn't set. So, setting it to OFF explicitly.")
+        print("INFO: MKL_FEATURE isn't set in the environment. So, setting it to OFF explicitly.")
         os.environ['MKL_FEATURE'] = "OFF"
--- a/open_ce/utils.py
+++ b/open_ce/utils.py
@@ -301,6 +301,13 @@ def detect_mkl_feature():
     This method detects if MKL_FEATURE flag is set in the environment to ON.
     If it is not set at all, the method sets it to OFF.
     '''
-    if not 'MKL_FEATURE' in os.environ:
+    if "MKL_FEATURE" not in os.environ:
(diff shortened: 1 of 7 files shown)


npanpaliya:
I'd forgotten to pass this variable as docker_build_args when build env is triggered with `--docker_build` option. I'll add that and update the PR.

open_ce/utils.py
```diff
@@ -295,3 +295,19 @@ def git_clone(git_url, git_tag, location):
         raise OpenCEError(Error.CLONE_REPO, git_url)
 
     return clone_successful
+
+def detect_mkl_feature():
```
bnemanich, line 299: It would probably be better to have a `--use_mkl` argument to the build tools, instead of relying on environment variables to be set by the user.

jayfurmanek: You think so? I think tool level arguments should be reserved for options that are more or less universal across the feedsocks. Other feedstocks could have some MKL specific function, but not many (maybe only PyTorch).

npanpaliya: I and Jason had discussed this over slack if we want to add environment variable or an argument to the build tools. 
@jayfurmanek and @bnemanich - Please let me know if we want to add an argument too.

bnemanich: If we are already adding logic in the tools to handle this specific environment variable, it would be better to use an argument. Another option would be to set whether to use MKL as part of the conda_build_config.yaml file. I think this might actually be the preferable way.

Turning things on and off with environment variables can make things harder to reproduce when running within multiple environments.

npanpaliya: I think adding the flag in conda_build_config.yaml of tensorflow-feedstock is appropriate. I've implemented it and tested, it went fine. With this, tools won't need any change at all. Just the frameworks which have some MKL specific functions would need to make the needed change in their conda_build_config.yaml and the respective build scripts.

npanpaliya:
Closing this one as the approach has been changed. Raised a new PR with new approach https://github.com/open-ce/tensorflow-feedstock/pull/33

PR #368, opened 2021-03-18
add test for image name

cdeepali:

## Description

Remove '-runtime' string being added to the image name. 
Also add a test for image name verification.

## Review process to land 

### DIFF
--- a/tests/build_image_test.py
+++ b/tests/build_image_test.py
@@ -201,5 +201,5 @@ def test_build_image_name(mocker):
     container_tool = utils.DEFAULT_CONTAINER_TOOL
 
     mocker.patch('os.system', return_value=0)
-    image_name = build_image.build_image("tests/testcondabuild", os.path.basename("tests/test-conda-env.yaml"), utils.DEFAULT_CONTAINER_TOOL)
+    image_name = build_image.build_image("tests/testcondabuild", os.path.basename("tests/test-conda-env.yaml"), container_tool)
     assert image_name == intended_image_name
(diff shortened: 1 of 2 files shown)


tests/build_image_test.py
```diff
@@ -91,7 +91,7 @@ def test_local_conda_channel_with_absolute_path(mocker):
     '''
     Test for build_runtime_image with local conda channel with its absolute path
     '''
-    intended_image_name = build_image.REPO_NAME + ":" + build_image.IMAGE_NAME
+    intended_image_name = build_image.REPO_NAME + ":" + build_image.IMAGE_NAME + "-" + "test-conda-env"
```
npanpaliya, line 94: Just a nit pick - Can we make this "test-conda-env" as constant variable which is declared once.

npanpaliya (approved):

cdeepali:
@jayfurmanek, @bnemanich, do we want to add arch to the image name as well ?

bnemanich:
@dllehr81 what do you think about the image name? I don't think we want to have the architecture name in it, but would like to verify.

tests/build_image_test.py
jayfurmanek: That's probably a good idea. That makes it easy to change if we need to.

cdeepali: I have added a commit to address this. @npanpaliya, @jayfurmanek please have a look.

cdeepali: Similar change would be needed in other tests too. I will create another PR for that.

npanpaliya (approved):
Looks good @cdeepali ! Thanks for fixing this all over.

bnemanich (approved):

cdeepali:
Merging this one.

PR #380, opened 2021-04-06
Add env files

bnemanich:

- [x] Did you update any affected [documentation](https://github.com/open-ce/open-ce/blob/main/doc/)?
- [x] Did you write any [tests](https://github.com/open-ce/open-ce/blob/main/tests/) to validate this change?  

## Description

Moves the Open-CE envs directory back into the open-ce repo.
Removes the build tools and associated documentation to the open-ce-builder repo.

## Review process to land 


README.md
```diff
@@ -4,168 +4,61 @@
...
-After performing the build using `open-ce build env`, the `open-ce build image` command can be used to create a runtime container image containing the newly created conda channel, as well as a conda environment with the newly build Open-CE packages. For more details on `open-ce build image`, please see [`doc/README.open_ce_build.md`](doc/README.open_ce_build.md#open-ce-build-image-sub-command).
+For more information on installing conda packages created from using the `open-ce` tools locally, please see the open-ce-builder [repository](https://github.com/open-ce/open-ce-builder).
 
-### Contributions
+## Contributions
 
 For contribution information, please see the [CONTRIBUTING.md](CONTRIBUTING.md) page.
 
-### Slack Community
+## Slack Community
```
jayfurmanek, line 62: Here we go. This is good here I think.

README.md
```diff
@@ -4,168 +4,61 @@
...
-git clone https://github.com/open-ce/open-ce-environments.git
-# Clone MY_FEEDSTOCK from GitHub
-git clone https://github.com/open-ce/MY_FEEDSTOCK-feedstock.git
-# Build packages
-./open-ce/open_ce/open-ce build feedstock --conda_build_config ./open-ce-environments/envs/conda_build_config.yaml --working_directory ./MY_FEEDSTOCK-feedstock
-```
+* `conda` >= 3.8.3
+  * The conda tool can either be installed through [Anaconda](https://www.anaconda.com/products/individual#Downloads) or [Miniconda](https://docs.conda.io/en/latest/miniconda.html).
 
 ### Installing Packages
```
jayfurmanek, line 41: Let's change this sections to ### Community Builders

We can also list MIT:
https://opence.mit.edu/#/

.github/actions/feedstock-optional-pr/action.yml
```diff
@@ -11,16 +11,15 @@ runs:
...
+        git clone https://github.com/open-ce/open-ce-builder.git \
+                  ./open-ce-builder
     - name: Run PR Tests
       shell: bash -l {0}
       run: |
-        test_path="$GITHUB_ACTION_PATH/../../../tests"
-        pip install -e "$GITHUB_ACTION_PATH/../../../"
-        python $test_path/feedstock_tests/check_build_numbers.py \
+        pip install -e open-ce-builer/open_ce
+        python open-ce-builder/tests/feedstock_tests/check_build_numbers.py \
```
bethune-bryant, line 23: Do we want to keep the `feedstock_tests` in the builder repo, or should we have that here?

.github/pull_request_template.md
```diff
@@ -2,7 +2,6 @@
 
 - [ ] Did you read the [contributor guide](https://github.com/open-ce/open-ce/blob/main/CONTRIBUTING.md)?
 - [ ] Did you update any affected [documentation](https://github.com/open-ce/open-ce/blob/main/doc/)?
-- [ ] Did you write any [tests](https://github.com/open-ce/open-ce/blob/main/tests/) to validate this change?  
 
 ## Description
```
bethune-bryant, line 7: We could probably remove thi sfile all together. The template [here](https://github.com/open-ce/.github/blob/main/pull_request_template.md) will be used if this file isn't here, and it's probably better to have that coming from one place.

.github/actions/feedstock-optional-pr/action.yml
bnemanich: I had it here at first, but when linting the file it complained about open-ce libraries not being found. I suppose we could install open-ce before calling pylint, but instead I moved it to the builder repo.

.github/pull_request_template.md
bnemanich: Ah, is that the same for the open-ce-builder one as well?

bethune-bryant: Yeah, any repo in the org will use what's in the `.github` repo if they don't have one defined. Ecen all the feedstocks.

bethune-bryant: I think maybe the one in open-ce just stayed there because it was there before the one in `.github`

.github/actions/feedstock-optional-pr/action.yml
bethune-bryant: With it only being one file, it's not that big a deal. If we ever need to make a PR in the feedstock-test action and in the open-ce-builder repo to fix something, maybe we move it then.

bethune-bryant (approved):

PR #381, opened 2021-04-06
[1.2] Pin updates for 1.2

jayfurmanek:

## Description

1.2 will include a update to protobuf across the board, which is a breaking change (hence the move up in versions)
ONNX and TensorFlow require patches to work.

TensorFlow Serving is temporarily removed until it supports newer protobuf upstream.

https://abi-laboratory.pro/?view=timeline&l=protobuf

closes: https://github.com/open-ce/open-ce-environments/pull/18

## Review process to land 

### DIFF
--- a/envs/arrow-env.yaml
+++ b/envs/arrow-env.yaml
@@ -1,2 +1,2 @@
 packages:
-  - feedstock : arrow-cpp
+  - feedstock : pyarrow
--- a/envs/arrow-env.yaml
+++ b/envs/arrow-env.yaml
@@ -1,2 +1,2 @@
 packages:
-￼  - feedstock : arrow-cpp
+  - feedstock : arrow-cpp
--- a/envs/arrow-env.yaml
+++ b/envs/arrow-env.yaml
@@ -1,0 +1,2 @@
+packages:
+￼  - feedstock : arrow-cpp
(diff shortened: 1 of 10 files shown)


envs/arrow-env.yaml
```diff
@@ -0,0 +1,2 @@
+packages:
+￼  - feedstock : arrow-cpp
```
npanpaliya, line 2: There is some garbage character over here which is also causing github checks to fail.

dncliss (approved):

envs/arrow-env.yaml
```diff
@@ -0,0 +1,2 @@
+packages:
+  - feedstock : arrow-cpp
```
bnemanich, line 2: Will the name of the feedstock be changed?

bnemanich:
Looks like tf-serving needs NCCL built as well.

bnemanich (approved):

envs/tensorflow-serving-env.yaml
```diff
@@ -0,0 +1,14 @@
+imported_envs:
```
npanpaliya, line 1: Should we have {% protobuf == "3.9.2" %} here as this environment file should not get executed for newer protobuf?

PR #382, opened 2021-04-07
change for default branch name

cdeepali:
Change for default branch name in the README.

### DIFF
--- a/doc/README.feedstock_setup.md
+++ b/doc/README.feedstock_setup.md
@@ -26,9 +26,6 @@ To enable CI tests, the [`feedstock-pr.yml` file](https://github.com/open-ce/.gi
 in the [Open-CE/.github repo](https://github.com/open-ce/.github)
 should be copied to the new feedstock repo at `.github/workflows/feedstock-pr.yml`
 
-On [line 5 of `feedstock-pr.yml`](https://github.com/open-ce/.github/blob/0958304e21dd7733e57dcaf3cbcf362dee877fbb/workflow-templates/feedstock-pr.yml#L5)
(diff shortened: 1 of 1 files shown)


bnemanich:
Why is this being removed? It seems like we want to keep this line in. We should probably change line 5 of feedstock-pr.yml from master to main though.

cdeepali:
Yes, that is already updated - https://github.com/open-ce/.github/blob/main/workflow-templates/feedstock-pr.yml#L5. So I thought this is not needed anymore. 

Do we still want to retain this?

jayfurmanek:
yeah - we fixed that last week with the PRs @dncliss pushed.

dncliss:
When I did the PR changes last week, I changed the wording in this section but I'd considered removing it entirely as @cdeepali does here.  I went ahead with the more minimal change just to make the review simpler for the mass-approval.
But really, I'm on board with this change.  Before the master/main switch, with some repos using one and a few using the other, a user might indeed have to edit this file to put the correct default branch name.  Now that all
repos have the same `main` default branch as reflected in this yaml file, there's no reason a human should need to edit the value, so it just becomes extraneous clutter information.

As per that opinion, I'm going to approve this -- but if there's not consensus on it then merging should probably wait.

dncliss (approved):
Approving; see my rationale in separate preceding comment.

bnemanich:
OK, sounds good to me.

bnemanich (approved):

cdeepali:
Thanks @dncliss, @bnemanich. merging this one.

PR #398, opened 2021-05-10
loosen pillow pin

jayfurmanek:

## Description

Pillow has move to 8.2.
Pillow 8.2 includes some needed security fixes that were not added to 8.1. 
Pillow 8.2 however does includes some API changes:
  https://github.com/python-pillow/Pillow/releases

This causes at least one torchvision test to fail:
  https://github.com/pytorch/vision/pull/3631

So far there doesn't seem to be any other impact to the frameworks. User code could still be affected so instead of moving the pin up to `8.2.*`, let's put it to `8.*.*` to allow 8.1 to still be used.

## Review process to land 


bnemanich (approved):

npanpaliya:
@jayfurmanek - I'm yet to confirm if TF works without pillow. My build of tensorflow-env.yaml is in progress. Once it is done, I will quickly run some tests as pillow is run time dependency of TF.

cdeepali (approved):

Issue #178, opened 2020-11-16
[FEEDSTOCK REQUEST] ONNX Runtime


jayfurmanek:
**Describe the package you'd like added**
ONNX Runtime is a cross-platform inferencing and training accelerator compatible with many popular ML/DNN frameworks, including PyTorch, TensorFlow/Keras, scikit-learn, and more. aka.ms/onnxruntime

Many users can benefit from ONNX Runtime, including those looking to:

Improve inference performance for a wide variety of ML models
Reduce time and cost of training large models
Train in Python but deploy into a C#/C++/Java app
Run on different hardware and operating systems
Support models created in several different frameworks

https://github.com/Microsoft/onnxruntime/releases/tag/v1.5.1

**Describe how this package fits in with the project**
ONNX Runtime inferencing is used by Nvidia's Triton Inferencing Server (used to be TensorRT inferencing server) among others.
It uses an plugs into many different cpu and gpu acceleration APIs

**Additional context**
ONNX Runtime inferencing APIs are stable and production-ready since the 1.0 release in October 2019 and can enable faster customer experiences and lower costs.

It's unclear if it works on anything other than x86

ONNX Runtime training feature was introduced in May 2020 in preview. This feature supports acceleration of PyTorch training on multi-node NVIDIA GPUs for transformer models. Additional updates for this feature are coming soon.

jayfurmanek:
CUDA11 support was added recently:
https://github.com/Microsoft/onnxruntime/releases/tag/v1.5.1

jayfurmanek:
I think this will line up with Open-CE 1.3:
 - onnx 1.7
 - CUDA support can be work in progress, but would be cuda11 variant, 
 - TRT support can be work in progress, but would be 7.1 (or later)
 - other noarch onnx tools to add as well:  
     - onnxmltools (https://github.com/onnx/onnxmltools), (1.7.x)
     - keras2onnx (https://github.com/onnx/keras-onnx/)  (1.7.x)
     - onnxconverter-common (https://github.com/microsoft/onnxconverter-common)  (1.7.x)

jayfurmanek:
tf2onnx seems up to date as well (https://github.com/onnx/tensorflow-onnx)

The latest version (1.8.5) added support for TF 2.5

jayfurmanek:
It looks like `onnxruntime` supports ranges of onnx opsets:

[...]

PR #450, opened 2021-07-07
Added pins for onnxruntime

npanpaliya:

## Description

Added pins needed for onnxruntime, needed for https://github.com/open-ce/open-ce/issues/178

## Review process to land 


envs/conda_build_config.yaml
```diff
@@ -18,6 +18,8 @@ bazel:
   - 3.7.*
 boost:
   - 1.67.0
+boost_mp11:
```
jayfurmanek, line 21: It looks like boost_mp11 is standalone from boost-cpp, but perhaps we'll need to consider bumping up boost-cpp as well.

envs/conda_build_config.yaml
```diff
@@ -80,13 +82,17 @@ google_pasta:
   - 0.2
 googleapis_common_protos:
   - 1.52.*
+gmock:
```
jayfurmanek, line 85: Do we need both `gmock` and `gtest`? I think they are the same thing these days.

jayfurmanek: It looks like they do still build into separate libraries but they are uniformly versioned. We could just use a single pin (`gtest`) for both test and mock I think - at the expense of a little bit of readability, but it's more proper - these two should stay linked.

envs/conda_build_config.yaml
```diff
@@ -112,6 +118,8 @@ keras_preprocessing:
   - 1.1.*
 leveldb:
   - 1.20.*
+libdate:
```
jayfurmanek, line 121: Maybe set this one to just 3.*
It looks like this is getting included into c++20, so it's likely there will only be 3.0.x updates at most with libdate, but in case there is a 3.1 let's allow that.

envs/conda_build_config.yaml
```diff
@@ -151,6 +159,8 @@ ninja:
   - 1.10.*
 nccl:
   - 2.8.*
+nlohmann_json:
```
jayfurmanek, line 162: It looks like this project marks things deprecated in x.y, but doesn't actually remove them until new x releases. We may be safe with 3.* here.

envs/conda_build_config.yaml
```diff
@@ -168,6 +178,8 @@ openjdk:
   - 1.8.0                      # [ppc64le]
 openmpi:
   - 3.1.3
+optional_lite:
```
jayfurmanek, line 181: ONNX sure loves their C++ extensions! For this one I don't really see any compatibility guidance.

envs/conda_build_config.yaml
npanpaliya: Yes, I too thought so but we will have to check DALI if it is fine with newer boost. So, I left that as it is for now until we update DALI.

envs/conda_build_config.yaml
npanpaliya: Okay.

envs/conda_build_config.yaml
npanpaliya: I kept it 3.0 because our recipe for libdate builds 3.0.1. In case there is a 3.1 version, we can update this too when we update the recipe for it. I'm okay with changing it to 3.*.

envs/conda_build_config.yaml
npanpaliya: Same thought I'd for this as I mentioned for libdate. But I'm fine with changing it to 3.*.

envs/conda_build_config.yaml
npanpaliya: Thank you ! :)

envs/onnx-env.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+packages:
+  - feedstock : safeint
+  - feedstock : gtest
+  - feedstock : libdate
+  - feedstock : onnx
+  - feedstock : optional-lite
+  - feedstock : boost_mp11
+  - feedstock : nlohmann_json
+  - feedstock : https://github.com/conda-forge/fire-feedstock
```
jayfurmanek, line 12: I think probably let's pick a commit here so its reproducible.

jayfurmanek (approved):

PR #501, opened 2021-08-27
Tf26 pins (and update to cos7 x86 compilers)

npanpaliya:

## Description

- Added pins needed for TF 2.6 and family refresh. 
- I've also included changes for new compilers. Using gcc 8.4 on x86 and 8.2 on ppc. So far I could verify TF env, xgboost (which was already using gcc 8), lightgbm with this PR.
- With these compiler updates, we don't need to pin libgcc-ng and libstdcxx-ng any more in the recipes. I'll push removal of those pins to the respective feedstock PRs after I verify all of them in x86/ppc and cuda 10.2/11.2 environments.
- We know DALI has build issues with gcc 8 which can be dealt separately either by fixing the build issues with gcc 8 or by falling back to its own compiler versions. I'll try to fix the build issues first.
- I'll do TF serving refresh and compiler changes together.

## Review process to land 


jayfurmanek:
Something may be malformed in the env file here. The PR check seems to not make sense.
The env file updates look ok though.

@bethune-bryant what do you think?

npanpaliya:
The PR check that is failing is because it is looking for keras-feedstock/recipe dir which doesn't exist yet as the recipe is yet to be added in main branch (PR is there for it).

envs/feedstock-patches/uwsgi/0001-disable-lto-adjust-pins.patch
```diff
@@ -59,12 +49,8 @@ index cf36a13..3016dc1 100644
      - yaml
 -    - zlib
 +    - zlib {{zlib }}
```
jayfurmanek, line 51: - zlib {{ zlib }}

jayfurmanek (approved):

jayfurmanek (approved):

jayfurmanek (approved):

PR #502, opened 2021-08-27
use list of supported CUDA arch in cuda_levels_details

cdeepali:

## Description

Magma code expects cuda capabilities to be passed as `sm_` options, so adding `sm_75` and `sm_80` for CUDA 10.2 and CUDA 11 respectively. 

Also using either `compute_75` or `compute_80` depending on CUDA version to support PTX for the latest supported CUDA arch.

## Review process to land 

### DIFF
--- a/envs/conda_build_config.yaml
+++ b/envs/conda_build_config.yaml
@@ -333,9 +333,9 @@ cuda_levels:
   - 3.7;5.2;6.0;7.0;7.5     # [x86_64]
   - 3.7;6.0;7.0;7.5         # [ppc64le]
 cuda_levels_details:
-  - sm_37,sm_52,sm_60,sm_61,sm_70,compute_75  # [x86_64]
-  - sm_37,sm_60,sm_70,compute_75              # [ppc64le]
+  - sm_37,sm_52,sm_60,sm_61,sm_70,sm_75,compute_75  # [x86_64]
+  - sm_37,sm_60,sm_70,sm_75,compute_75              # [ppc64le]
 cuda11_levels:
   - 8.0
 cuda11_levels_details:
-  - compute_80
+  - sm_80,compute_80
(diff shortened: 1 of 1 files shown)


envs/conda_build_config.yaml
```diff
@@ -333,9 +333,9 @@ cuda_levels:
   - 3.7;5.2;6.0;7.0;7.5     # [x86_64]
   - 3.7;6.0;7.0;7.5         # [ppc64le]
 cuda_levels_details:
-  - sm_37,sm_52,sm_60,sm_61,sm_70,compute_75  # [x86_64]
-  - sm_37,sm_60,sm_70,compute_75              # [ppc64le]
+  - sm_37,sm_52,sm_60,sm_61,sm_70,sm_75,compute_75        # [x86_64 and cudatoolkit == '10.2']
```
bnemanich, line 336: We should probably merge `cuda11_levels` with `cuda_levels` as well. However, you might need to leave in `cuda11_levels_details` for now so that it doesn't break TF. I think you will need to merge in the new cuda_levels_details, then fix TF (and anything else that use `cuda11_levels_details`), then remove `cuda11_levels_details`.

cdeepali: Yes, I have been thinking so too regarding the `cuda_levels`. Will add that change.

cdeepali: I have added a PR for TF - https://github.com/open-ce/tensorflow-feedstock/pull/65 to fix this.

jayfurmanek: yeah, this is a better idea.

jayfurmanek (approved):

jayfurmanek:
I pushed an update to switch from semi-colons to commas.

envs/conda_build_config.yaml
cdeepali: If we merge TF PR, along with this PR, I think we would not break TF build.

PR #507, opened 2021-08-31
pins for transformers v4.9.2

cdeepali:

## Description

Add pins for transformers v4.9.2. 

`Sentencepiece` is an optional dependency of transformers, but some tests require `sentencepiece`. Adding `sentencepiece` to the `transformers-env.yaml` makes it available in the generated environment yaml which is later used for creating the test environment. 

Some tests also require `pyarrow`, so building it too, as part of the `transformers-env.yaml`. 

`pyarrow` available on defaults could not be used because the following confict:
**pyarrow on defaults**:-> `arrow-cpp ->    libprotobuf >=3.11.2,<3.12.0a0`
**sentencepiece on open-ce**:-> `  protobuf 3.14.*`

## Review process to land 

### DIFF
--- a/envs/conda_build_config.yaml
+++ b/envs/conda_build_config.yaml
@@ -228,7 +228,7 @@ regex:
 rust:
   - 1.54.0
 rust_compiler_version:
-- '1.46'
+  - '1.46'
 safeint:
   - 3.0
 scipy:
--- a/envs/conda_build_config.yaml
+++ b/envs/conda_build_config.yaml
@@ -55,7 +55,7 @@ cymem:
 cython:
   - 0.29
 cython_blis:
-  - 0.4.*
+  - 0.7.*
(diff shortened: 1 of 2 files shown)


jayfurmanek:
Can you add here some of the details we discussed on why this is needed. (some of the tests need sentencepiece/arrow)

jayfurmanek (approved):

envs/conda_build_config.yaml
```diff
@@ -225,7 +225,7 @@ requests:
 regex:
   - 2020.11.*
 rust_nightly:
```
jayfurmanek, line 227: we may need to just switch this to just `rust`

cdeepali:
Requires - https://github.com/open-ce/pyarrow-feedstock/pull/19.

PR #522, opened 2021-09-21
remove arrow-env as this is no longer needed

cdeepali:

## Description

This PR removes `arrow-env` from `transformers-env.yaml`. arrow-env was added earlier because `pyarrow` was required by datasets package which was installed using pip in the `open-ce-tests.yaml`.

However, the entire transformers test suite is taking more than 3 hours, so have raised a PR - https://github.com/open-ce/transformers-feedstock/pull/24 to execute only a small subset of the test suite as part of the `open-ce-tests.yaml`. So datasets and hence arrow is not required for test execution. 

## Review process to land 

### DIFF
--- a/envs/transformers-env.yaml
+++ b/envs/transformers-env.yaml
@@ -1,5 +1,3 @@
-imported_envs:
-  - arrow-env.yaml
 packages:
   - feedstock : tokenizers
   - feedstock : huggingface_hub


alexlang74 (approved):
Thanks for catching that, that's a nice "side effect"

npanpaliya (approved):

PR #574, opened 2022-01-03
Fixed pyarrow build on s390x

npanpaliya:

## Description

Fixes pyarrow build on s390x. The failure was due to orc which was being built using the latest protobuf  which is 3.19.

## Review process to land 

### DIFF
--- a/envs/arrow-env.yaml
+++ b/envs/arrow-env.yaml
@@ -6,6 +6,8 @@ packages:
       - feedstock-patches/rapidjson/s390x-fix1.patch
   - feedstock : https://github.com/AnacondaRecipes/orc-feedstock
     git_tag: 47b595ae87b7bd10b16e08aef41e1b9e9d257848
+    patches:
+      - feedstock-patches/orc/0001-Pin-protobuf-to-3.14.patch
   - feedstock : https://github.com/conda-forge/libutf8proc-feedstock
     git_tag: d2a2366dc3c02405b1b54b991758b5bb57b75e6d
   - feedstock : https://github.com/AnacondaRecipes/re2-feedstock
--- a/envs/feedstock-patches/orc/0001-Pin-protobuf-to-3.14.patch
+++ b/envs/feedstock-patches/orc/0001-Pin-protobuf-to-3.14.patch
@@ -1,0 +1,31 @@
+From 8dbdb28766946ddd90ba4490984a797759b25666 Mon Sep 17 00:00:00 2001
+From: Nishidha Panpaliya <npanpa23@in.ibm.com>
+Date: Mon, 3 Jan 2022 13:59:25 +0000
+Subject: [PATCH] Pin protobuf to 3.14
+
+---
+ recipe/meta.yaml | 4 ++--
+ 1 file changed, 2 insertions(+), 2 deletions(-)
+
+diff --git a/recipe/meta.yaml b/recipe/meta.yaml
+index f4dd818..1b649f9 100644
+--- a/recipe/meta.yaml
++++ b/recipe/meta.yaml
+@@ -28,12 +28,12 @@ requirements:
+     - patch  # [not win]
+     - ninja       # [unix]
+     # `protoc` is also used for building
+-    - libprotobuf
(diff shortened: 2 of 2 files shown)


envs/arrow-env.yaml
```diff
@@ -27,10 +29,10 @@ packages:
   - feedstock : https://github.com/AnacondaRecipes/thrift-cpp-feedstock
     git_tag: 09e3838f8caaa3ab9728b111620422dd524459e1
 {% endif %}
-{% if build_type == 'cuda' %}
+{% if build_type == 'cuda' and not s390x %}
   - feedstock : cudatoolkit    #[cudatoolkit == "11.2"]
 {% endif %}
-  - feedstock : grpc-cpp
-  - feedstock : pyarrow
+  - feedstock : grpc-cpp       #[not s390x or build_type == 'cpu']
```
alexlang74, line 35: This is `(NOT s 390x) OR (build_type = CPU)` , correct? I was wondering about the scope of the not (as it _could_ also be `NOT (s390 OR type=cpu)` )

alexlang74 (approved):

envs/arrow-env.yaml
npanpaliya: This means grpc-cpp is to be built for all platforms except s390x or if it is s390x, then build only for `cpu`.

ravigumm: Stepping back a bit to understand why this is an issue now and not earlier.
Were we sending --build-type cpu as a command line argument earlier and not now ?
Shall we (1) set build_type in conda build config yaml file as below ?
```
build_type:
  - cpu
  - cuda # [not s390x]
```
OR
pass build-type command line option for s390x builds ?

envs/feedstock-patches/orc/0001-Pin-protobuf-to-3.14.patch
```diff
@@ -0,0 +1,31 @@
...
+     - ninja       # [unix]
+     # `protoc` is also used for building
+-    - libprotobuf
++    - libprotobuf {{ protobuf }}
+     - cmake
+   host:
+     - zlib
+     - snappy
+-    - libprotobuf
++    - libprotobuf {{ protobuf }}
```
ravigumm, line 25: This LGTM

envs/opence-env.yaml
```diff
@@ -10,7 +10,7 @@ imported_envs:
   - transformers-env.yaml
   - horovod-env.yaml
   - lightgbm-env.yaml
-  - arrow-env.yaml
+  - arrow-env.yaml   #[not s390x or build_type == 'cpu']
```
ravigumm, line 13: Is this condition required here ? Just for my understanding, Are we trying to run envs/opence-env.yaml on s390x ? Then do we have to mark all other -env.yaml lines in this file with [not s390x] ?

envs/arrow-env.yaml
ravigumm: In the upcoming months, we will have tensorflow-env.yaml, tensorboard-env.yaml, etc enabled for s390x also and then those files also will require these conditions inside those files. My intention would be to avoid some of these these conditions in each -env.yaml file, if makes sense.

envs/opence-env.yaml
npanpaliya: We don't usually run opence-env.yaml on s390x. So, you're right! Either we should add [not s390x] for all the envs entries or remove this change although keeping this change won't harm as such.

cdeepali: Probably we can avoid having condition here to keep this yaml simple.

envs/arrow-env.yaml
npanpaliya: We discussed this on slack and in our scrum and decided that we will revert these conditions for now. CI systems are anyway running s390x builds for build_type cpu. And in open-ce 1.6, we can tackle this by blocking cuda build type for s390x in builder code itself.

cdeepali (approved):

ravigumm (commented):
Discussed with Nishidha offline. LGTM. --build-type command line option to be used for s390x.

PR #606, opened 2022-04-11
clangdev fix and more s390x deps

ravigumm:

## Description
clangdev fix for s390x and more dependencies of tensorflow for s390x

Fixes # (issue).

## Review process to land 

### DIFF
--- a/envs/tensorflow-s390x-extra-deps-env.yaml
+++ b/envs/tensorflow-s390x-extra-deps-env.yaml
@@ -2,4 +2,12 @@ packages:
 {% if s390x %}
   - feedstock : https://github.com/AnacondaRecipes/flatbuffers-feedstock.git
     git_tag: c7bd7770705d34151e80cae10b7b4a4cc960094f
+  - feedstock : https://github.com/conda-forge/llvmdev-feedstock
+    git_tag: ddce3b23502e4a4c97650062c709014c22144183
+  - feedstock : https://github.com/conda-forge/portpicker-feedstock
+    git_tag: b21a22eb2c31f135b8e2966b0e53c269e9a12a18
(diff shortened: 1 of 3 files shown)


ravigumm:
@npanpaliya , @cdeepali , @pradghos, Would you please review ? Thanks

envs/tensorflow-s390x-extra-deps-env.yaml
```diff
@@ -2,4 +2,12 @@ packages:
 {% if s390x %}
   - feedstock : https://github.com/AnacondaRecipes/flatbuffers-feedstock.git
     git_tag: c7bd7770705d34151e80cae10b7b4a4cc960094f
+  - feedstock : https://github.com/conda-forge/llvmdev-feedstock
+    git_tag: ddce3b23502e4a4c97650062c709014c22144183
+  - feedstock : https://github.com/conda-forge/portpicker-feedstock
+    git_tag: b21a22eb2c31f135b8e2966b0e53c269e9a12a18
+  - feedstock : https://github.com/AnacondaRecipes/googleapis-common-protos-feedstock
```
pradghos, line 9: @ravigumm : `portpicker` , `googleapis-common-protos`  , `promise` or in general - I think it would be better to pin the packages like - `python`,  `protobuf` or other packages for run time dependency.

npanpaliya (approved):

envs/tensorflow-s390x-extra-deps-env.yaml
ravigumm: As there are no changes to the feedstock code done by us and because we will anyway pass python version from command line, we can keep the remaining packages versions flexible similar to other open-ce feedstocks --- keeping this yaml file simpler.

pradghos: Sure, Thanks!

PR #672, opened 2022-08-16
Added new opence-s390x-env.yaml for s390x build

pradghos:

## Description
- introduced opence-s390x-env.yaml to manage s390x builds

## Review process to land 

### DIFF
--- a/envs/opence-s390x-env.yaml
+++ b/envs/opence-s390x-env.yaml
@@ -1,6 +1,6 @@
 imported_envs:
 {% if build_type == 'cpu' %}
-    - tensorflow-env.yaml
+  - tensorflow-env.yaml
 {% endif %}
   - xgboost-env.yaml
   - arrow-env.yaml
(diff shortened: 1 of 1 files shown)


envs/opence-s390x-env.yaml
```diff
@@ -0,0 +1,9 @@
+imported_envs:
+{% if build_type == 'cpu' %}
```
npanpaliya, line 2: I think this file itself be complete for s390x if someone wants to build it using external repos. Is this the final list of envs that we build on s390x?

envs/opence-s390x-env.yaml
```diff
@@ -0,0 +1,9 @@
...
+    - tensorflow-env.yaml
```
npanpaliya, line 3: This should be at the same indentation as that of other envs. Because when this build_type condition will be resolved, the extra indented `tensorflow-env.yaml` won't be read correctly and this will give you an error exactly same as that of in the PR check.

npanpaliya (approved):

PR #707, opened 2022-10-10
Import p10 conda_build_config in opence-p10-env file

npanpaliya:

## Description

This will remove the need of explicit argument specification for P10 builds. One just has to specify `--ppc_arch-p10` for P10 builds after this change.

## Review process to land 

### DIFF
--- a/envs/opence-p10-env.yaml
+++ b/envs/opence-p10-env.yaml
@@ -1,5 +1,9 @@
 builder_version: ">=10.0.1"
 
+conda_build_configs:
+  - conda_build_config.yaml
+  - conda_build_config_p10.yaml
+
 imported_envs:
   - openblas-env.yaml
   - onnx-env.yaml


npanpaliya:
> Good idea! You'll update the docs in a separate PR, I assume? Do the CI workflows need to change?

With document update, do you mean this https://github.com/open-ce/open-ce-builder/blob/main/doc/README.open_ce_build.md#build-packages? If yes, I think this doesn't need to be changed much, as with individual env files, we still need to provide `--conda_build_config=./open-ce/envs/conda_build_config.yaml,./open-ce/envs/conda_build_config_p10.yaml` argument. I'll still update this doc a bit saying with `opence-p10-env.yaml` one can skip this argument.
We need to check CI workflows. I'll let them know about this change, so that they can find out if the workflows need any changes.

npanpaliya:
@ketank-new

cdeepali:
@npanpaliya, Can similar change also go in `opence-env.yaml`?

npanpaliya:
For opence-env.yaml, it doesn't matter if it is present or not. Builder handles that by default. For P10 builds also, we could have added code in builder to import `conda_build_config_p10.yaml` but this change is better than hard-coding in code.

cdeepali (approved):

PR #737, opened 2022-12-12
Fixed install time conflict issue on s390x

npanpaliya:

## Description

The issue was coming due to `mpi 1.0 mpich` entry in conda env file whereas openmpi package was built with `mpi 1.0 openmpi`. This was causing install time conflict while conda env creation. So, as a fix, we have disabled mpi build for `mpich` variant. 

Ideally `mpi 1.0 mpich` and `mpi 1.0 openmpi` both entries should have been written in conda env files as mpi package for both variants is built in the output folder. But builder code overwrites `mpi 1.0 openmpi` with `mpi 1.0 mpich` assuming both are same (build string is not accounted while comparing packages to be written into conda env files). Also, this issue recreation depends on the order of the packages being traversed by build-tree. So, it may not reproduce every time. Hence, we also need to address this kind of scenario in builder code itself, which I'll check later. 

## Review process to land 


ravigumm (approved):
LGTM. We can explore mpich later

cdeepali (approved):

PR #795, opened 2023-04-17
fix build on PPC with cuda11.2

cdeepali:

## Description

Fixes # (issue).

## Review process to land 


alexlang74 (approved):

npanpaliya:
This patch was not needed on x86 and hence I'd removed it from all the places. It's missed from being removed at one place in transformers-env.yaml. So, we need to remove that as well. But please check the build on ppc once without this patch.

npanpaliya (approved):

cdeepali:
Okay, updated to remove the patch. The build went fine locally.

Issue #1349, opened 2025-04-16
[BUG] transformer feedstock failure on Z


smartvibs8876:
Transformers 4-37-2 and 4-38-0 are failing on z

```
Cloning into bare repository '/opt/conda/conda-bld/git_cache/github.com/huggingface/transformers'...
[d9eee5f23d94405d90b7e9ff88b9443fee42f8528a658f54214c2aba7530d80c] Object does not exist on the server: [404] Object does not exist on the server
error: failed to fetch some objects from 'https://github.com/huggingface/transformers.git/info/lfs'
fatal: destination path '/opt/conda/conda-bld/git_cache/github.com/huggingface/transformers' already exists and is not an empty directory.

```

cdeepali:
Fixed by #1351

PR #1351, opened 2025-04-23
Resolve s390x transformer build issue -  lfs object not found

smartvibs8876:
Fixes issue #1349 

The issue with transformers build on s390x is with cloning of source code . It make use of LFS storage to store certain files. This LFS storage is removed from the branch 4.38.0 and 4.37.2 we are building which we confirmed using following commands.

```
(base) [builder@f1e50114fef8 transformers]$ git lfs ls-files
(base) [builder@f1e50114fef8 transformers]$ cat .gitattributes
*.py    eol=lf
*.rst   eol=lf
*.md    eol=lf
*.mdx   eol=lf

```

conda internally executes this and fails
```
Command '['/opt/conda/bin/git', 'lfs', 'fetch', 'origin', '--all']' returned non-zero exit status 2.
```

Solution:-Change source to use tar.gz from github instead of cloning source code


envs/transformers-env.yaml
```diff
@@ -28,7 +28,7 @@ packages:
   - feedstock : https://github.com/conda-forge/transformers-feedstock
     git_tag : 053ca66df0d4963ad295b5e2f9f0d0a2f42c51b1
     patches :
-      - feedstock-patches/transformers/0001-Fixed-CVE-2024-11392-11393-11394-12720.patch
+      - feedstock-patches/transformers/0001-Fixed-CVE-2024-11392-11393-11394-12720-Source-Change.patch
```
cdeepali: You can continue with same name for patch file.

smartvibs8876: Patch renamed to its original name

smartvibs8876:
This PR is to resolve issue mentioned #1349 

The issue with transformers build on s390x is with cloning of source code . It make use of LFS storage to store certain files. This LFS storage is removed from the branch 4.38.0 and 4.37.2 we are building which we confirmed using following commands.
```
(base) [builder@f1e50114fef8 transformers]$ git lfs ls-files
(base) [builder@f1e50114fef8 transformers]$ cat .gitattributes
*.py    eol=lf
*.rst   eol=lf
*.md    eol=lf
*.mdx   eol=lf
```

conda  internally executes this and fails
```
Command '['/opt/conda/bin/git', 'lfs', 'fetch', 'origin', '--all']' returned non-zero exit status 2.
```

Solution:-Change source to use tar.gz from github instead of cloning source code

cdeepali (approved):

